PR #2, opened 2018-06-23
Checkout repos at latest tag by default

jeanconn:
Instead of checking out just the tip of master, checkout the most recent tag.

### DIFF
--- a/ska_conda/ska_builder.py
+++ b/ska_conda/ska_builder.py
@@ -17,26 +17,48 @@ class SkaBuilder(object):
         self.ska_src_dir = os.path.join(self.ska_root, "src")
         os.environ["SKA_TOP_SRC_DIR"] = self.ska_src_dir
 
-    def _clone_repo(self, name):
+    def _clone_repo(self, name, tag=None):
         if name == "ska":
             return
         print("Cloning source %s." % name)
         clone_path = os.path.join(self.ska_src_dir, name)
-        if os.path.exists(clone_path):
-            print("Source %s exists, skipping." % name)
-            return
-        yml = os.path.join(pkg_defs_path, name, "meta.yaml")
-        with open(yml) as f:
-            requires = False
-            while not requires:
-                line = f.readline().strip()
-                if line.startswith("path:"):
-                    requires = True
-            data = yaml.load(f)
-            url = data['about']['home']
-            git.Repo.clone_from(url, clone_path)
+        if not os.path.exists(clone_path):
+            # Try ssh first to avoid needing passwords for the private repos
+            # We could add these ssh strings to the meta.yaml for convenience
+            try:
+                git_ssh_path = 'git@github.com:sot/' + name + '.git'
+                repo = git.Repo.clone_from(git_ssh_path, clone_path)
+                assert not repo.bare
+            except:
+                yml = os.path.join(pkg_defs_path, name, "meta.yaml")
+                with open(yml) as f:
+                    requires = False
+                    while not requires:
+                        line = f.readline().strip()
+                        if line.startswith("path:"):
+                            requires = True
+                    data = yaml.load(f)
+                    url = data['about']['home']
+                repo = git.Repo.clone_from(url, clone_path)
+        else:
+            repo = git.Repo(clone_path)
+        assert not repo.is_dirty()
+        # I think we want the commit/tag with the most recent date, though
+        # if we actually want the most recently created tag, that would probably be
+        # tags = sorted(repo.tags, key=lambda t: t.tag.tagged_date)
+        # I suppose we could also use github to get the most recent release (not tag)
+        if tag is None:
+            tags = sorted(repo.tags, key=lambda t: t.commit.committed_datetime)
+            repo.git.checkout(tags[-1].name)
+            if tags[-1].commit == repo.heads.master.commit:
+                print("Auto-checked out at {} which is also tip of master".format(tags[-1].name))
+            else:
+                print("Auto-checked out at {} NOT AT tip of master".format(tags[-1].name))
+        else:
+            repo.git.checkout(tag)
+            print("Checked out at {}".format(tag))
 
-    def clone_one_package(self, name):
+    def clone_one_package(self, name, tag=None):
         self._clone_repo(name)
 
     def clone_all_packages(self):
--- a/ska_conda/ska_builder.py
+++ b/ska_conda/ska_builder.py
@@ -11,7 +11,7 @@ class SkaBuilder(object):
 
     def __init__(self, ska_root=None):
         if ska_root is None:
-            ska_root = "/data/acis/ska3_pkg/"
+            ska_root = "/tmp/ska3_pkg/"
         self.ska_root = ska_root
         self.ska_build_dir = os.path.join(self.ska_root, "builds")
         self.ska_src_dir = os.path.join(self.ska_root, "src")


jeanconn:
For an idea of which packages need some tag updates, here's the output with my silly print statements.

```
In [1]: run scripts/clone_ska_sources all
Cloning source Ska.Shell.
Auto-checked out at 3.3.1 NOT AT tip of master
Cloning source Ska.File.
Auto-checked out at 3.4.1 NOT AT tip of master
Cloning source pyyaks.
Auto-checked out at 3.3.4 NOT AT tip of master
Cloning source ska_path.
Auto-checked out at 3.1 NOT AT tip of master
Cloning source testr.
Auto-checked out at 3.2 which is also tip of master
Cloning source Ska.tdb.
Auto-checked out at 3.5.1 NOT AT tip of master
Cloning source Chandra.Time.
Auto-checked out at 3.20.1 NOT AT tip of master
Cloning source Ska.ParseCM.
Auto-checked out at 3.3.1 which is also tip of master
Cloning source Ska.DBI.
Auto-checked out at 3.8.2 NOT AT tip of master
Cloning source Ska.ftp.
Auto-checked out at 3.4.3 NOT AT tip of master
Cloning source Ska.Numpy.
Auto-checked out at 3.8.1 NOT AT tip of master
Cloning source Quaternion.
Auto-checked out at 3.4.1 which is also tip of master
Cloning source Ska.engarchive.
Auto-checked out at 3.43 NOT AT tip of master
Cloning source kadi.
Auto-checked out at 3.15.2 NOT AT tip of master
Cloning source Ska.Matplotlib.
Auto-checked out at 3.11.2 NOT AT tip of master
Cloning source Ska.quatutil.
Auto-checked out at 3.3.1 NOT AT tip of master
Cloning source Ska.Sun.
Auto-checked out at 3.5 which is also tip of master
Cloning source Chandra.Maneuver.
Auto-checked out at 3.7 which is also tip of master
Cloning source cmd_states.
Auto-checked out at 3.14 which is also tip of master
Cloning source xija.
Auto-checked out at 3.9 which is also tip of master
Cloning source maude.
Auto-checked out at 3.1 NOT AT tip of master
```

ska_conda/ska_builder.py
```diff
@@ -11,32 +11,54 @@ class SkaBuilder(object):
...
+        else:
+            repo = git.Repo(clone_path)
+        assert not repo.is_dirty()
+        # I think we want the commit/tag with the most recent date, though
+        # if we actually want the most recently created tag, that would probably be
+        # tags = sorted(repo.tags, key=lambda t: t.tag.tagged_date)
+        # I suppose we could also use github to get the most recent release (not tag)
+        if tag is None:
+            tags = sorted(repo.tags, key=lambda t: t.commit.committed_datetime)
+            repo.git.checkout(tags[-1].name)
```
jeanconn: This is obviously just one way of going about this.  I'm also wondering now if the {{ GIT_DESCRIBE_TAG }} method can break if the tag isn't formatted in a way that conda will like.  I note their docs say "Conda acknowledges PEP 440" but I don't really know what acknowledges means in that context.  We can obviously just update tags to work if they are broken, so that's just something to keep in mind.

taldcroft:
This is going in the right direction.  Overall I feel like the original implementation and API can be simplified.  @jzuhone maybe had some different uses in mind, but I feel like there basically just needs to be three methods, `__init__`, `_get_repo(self, name, tag=None)`, and `build_packages(self, name=None, tag=None)`.

`_get_repo` either clones a fresh copy or uses the existing repo, and takes care of the work of fetching / pulling to make sure the return `repo` is at the correct tag (or latest if no `tag` is supplied).  I don't immediately see a need an API for separately cloning / updating, that just happens as a matter of course when building.

`build_packages` builds either all the packages (if `name is None`) or builds the specified package.  This can be done neatly with:
```
if name is None:
    names = [nm.strip() for nm in open(BUILD_LIST, "r") if nm.strip() and not nm.startswith('#')]
else:
    names = [name]
for name in names:
    repo = self._get_repo(name, tag)
    # Now do the few steps for building `repo`, probably checking to see if
    # the expected output conda package is already there.  This code is probably
    # short enough to just be in the loop here, but if not could be factored out to
    # self._build_package(repo)
```

ska_conda/ska_builder.py
```diff
@@ -11,32 +11,54 @@ class SkaBuilder(object):
 
     def __init__(self, ska_root=None):
```
taldcroft: Add `user='sot', git_repo_path='git@github.com:{user}/{name}.git'` kwargs (and set corresponding instance attrs).

ska_conda/ska_builder.py
```diff
@@ -11,32 +11,54 @@ class SkaBuilder(object):
...
-                if line.startswith("path:"):
-                    requires = True
-            data = yaml.load(f)
-            url = data['about']['home']
-            git.Repo.clone_from(url, clone_path)
+        if not os.path.exists(clone_path):
+            # Try ssh first to avoid needing passwords for the private repos
+            # We could add these ssh strings to the meta.yaml for convenience
+            try:
+                git_ssh_path = 'git@github.com:sot/' + name + '.git'
```
taldcroft: `repo = git.Repo.clone_from(self.git_repo_path.format(user=self.user, name=name), clone_path)`

taldcroft:
With the changes here, it seems this is not so hardwired to building Ska and instead it is more about generally maintaining a list of conda packages.  If you move these into the `SkaBuilder` init then this can really be used to build anything given a directory of build recipes and a spec of the package order.
```
pkg_defs_path = os.path.join(ska_conda_path, "pkg_defs")
build_list = os.path.join(ska_conda_path, "build_order.txt")
```

ska_conda/ska_builder.py
```diff
@@ -11,32 +11,54 @@ class SkaBuilder(object):
...
+                    requires = False
+                    while not requires:
+                        line = f.readline().strip()
+                        if line.startswith("path:"):
+                            requires = True
+                    data = yaml.load(f)
+                    url = data['about']['home']
+                repo = git.Repo.clone_from(url, clone_path)
+        else:
+            repo = git.Repo(clone_path)
```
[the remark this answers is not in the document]
jeanconn: Good point!  Though I'm wondering if we need to think about the use cases and problems of  the "already-existing" dir some more anyway .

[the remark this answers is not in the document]
taldcroft: We stipulate that the existing directory is created and maintained only via this script.  Problems then seem extremely unlikely.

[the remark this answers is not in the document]
jeanconn: OK, so do we need a way to build a custom/test package?  Or would we just do that outside this tool?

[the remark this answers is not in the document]
taldcroft: I think we should be able to handle this procedurally by making a release candidate tag such as `3.43.4rc1`.  Apart from testing the packaging itself, one can do functional/integration testing using pip install from the development git repo.  I.e. spin up a dev Ska3, conda uninstall the package in question, pip install the dev version, then test.  Or conda install using the local build path.  Since "conda acknowledges PEP440", this *should* work.

But let's defer further specifics of this use case for a future PR if needed, it will be easier based on an already-working system.  I'll stub a placeholder into the process wiki.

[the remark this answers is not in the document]
jeanconn: So adding `get fetch origin` seems fine if you want to reuse the repos dirs over some period of time and want to make sure you have the newest tags at origin.  We could also just clone the repos fresh for every run of the tool and remove this code path, but that does seem a little annoying.

taldcroft:
> For an idea of which packages need some tag updates, here's the output with my silly print statements.

It turns out most of those `NOT AT tip of master` are because of the sot-wide update to add license lines to code files.  I have done a scrub of that list and added appropriate release tags to the four actual cases where the tag doesn't match what is in skare/pkgs.manifest (py3).

ska_conda/ska_builder.py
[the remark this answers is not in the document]
taldcroft: Yes, I envision an "official" build process for production, done as `aca` user, that uses a permanent location like `/proj/sot/ska3/conda-builds`, or something like that.  Fetching and pulling is definitely faster than re-cloning the whole thing every time.

jeanconn:
At some point we probably also need to figure out build(er) requirements.  ska_builder is presently using the `git` Python module which I don't think is in our current vanilla/non-ska conda env.

taldcroft:
> At some point we probably also need to figure out build(er) requirements. ska_builder is presently using the git Python module which I don't think is in our current vanilla/non-ska conda env.

No worries.  This will be captured in the process doc.

ska_conda/ska_builder.py
```diff
@@ -11,32 +11,56 @@ class SkaBuilder(object):
...
+            repo.remotes.origin.fetch()
+            repo.remotes.origin.fetch("--tags")
```
jeanconn: As far as I could tell, the fetching behavior seems to differ a bit based on git version, and I'm not sure about gitpython at all.  Using both fetch and fetch --tags seemed, at worse, duplication.
I think that fetching from 'origin' will be appropriate in all cases.

taldcroft: We can use conda git for this right?  That should make behavior uniform.  Git fetch origin alone will always get the tags.

jeanconn: I think gitpython gets the git in your path.  And yes, conda git should probably work as the thing it gets in the path.  Should get added to requirements (and I don't recall which environments/installs had issues with https).     Just when trying to figure out how you do a fetch with gitpython, I had also seen the language that "most" tags should be reachable via `git fetch` and hadn't figured the exceptions yet.

ska_conda/ska_builder.py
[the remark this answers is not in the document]
jeanconn: I went with `build_root` instead of `conda_build_root` but can obviously just edit again.

ska_conda/ska_builder.py
```diff
@@ -59,30 +81,20 @@ def _build_package(self, name):
         print("Building package %s." % name)
         pkg_path = os.path.join(pkg_defs_path, name)
         cmd_list = ["conda", "build", pkg_path, "--croot",
-                    self.ska_build_dir, "--no-test",
-                    "--no-anaconda-upload"]
+                    self.build_dir, "--no-test",
+                    "--no-anaconda-upload", "--skip-existing"]
```
jeanconn: It looks like the --skip-existing option is smart enough that it skips if you have an existing build at the requested version (not just a package built with the name) so this seems to just work for our use cases.

ska_conda/ska_builder.py
```diff
@@ -59,30 +81,20 @@ def _build_package(self, name):
...
         subprocess.run(cmd_list)
 
     def build_one_package(self, name):
         repo = self._get_repo(name)
-        if repo is not None:
-            repo.remote().pull()
```
jeanconn: I cut the pulls and the pull version checks below in favor of that --skip-existing option.

ska_conda/ska_builder.py
```diff
@@ -59,30 +81,20 @@ def _build_package(self, name):
         print("Building package %s." % name)
         pkg_path = os.path.join(pkg_defs_path, name)
         cmd_list = ["conda", "build", pkg_path, "--croot",
-                    self.ska_build_dir, "--no-test",
-                    "--no-anaconda-upload"]
+                    self.build_dir, "--no-test",
```
jeanconn: It may take a little longer to build, but I'd prefer to fix the packages as needed so we can actually run the tests if provided.

ska_conda/ska_builder.py
```diff
@@ -9,34 +9,56 @@
...
-            ska_root = "/data/acis/ska3_pkg/"
-        self.ska_root = ska_root
-        self.ska_build_dir = os.path.join(self.ska_root, "builds")
-        self.ska_src_dir = os.path.join(self.ska_root, "src")
-        os.environ["SKA_TOP_SRC_DIR"] = self.ska_src_dir
+    def __init__(self, build_root='.', user='sot', git_repo_path='git@github.com:{user}/{name}.git'):
+        self.user = user
+        self.git_repo_path = git_repo_path
+        self.build_dir = os.path.abspath(os.path.join(build_root, "builds"))
+        self.src_dir = os.path.abspath(os.path.join(build_root, "src"))
```
jeanconn: I don't know if there would be any other issues using abspath when trying to work somewhat relatively (using a build root explicitly defined as "." or something else relative), but this seems to work.

ska_conda/ska_builder.py
```diff
@@ -9,34 +9,55 @@
...
+        os.environ["SKA_TOP_SRC_DIR"] = self.src_dir
```
jeanconn: Not sure if we want to use this as a somewhat global environment variable, or pass it to conda build as an env in the subprocess.

taldcroft:
Do you see any need for the clone_ska_sources script?  I don't.

ska_conda/ska_builder.py
taldcroft: I think it doesn't matter.

ska_conda/ska_builder.py
taldcroft: Can you elaborate, I don't understand this comment.

jeanconn:
It was helpful in testing/modifying the process to have the script to separate the two tasks but can probably be safely removed when we're done.

ska_conda/ska_builder.py
taldcroft: Yes, pull was never needed as long as you do a fetch and subsequent checkout at the desired tag.

ska_conda/ska_builder.py
taldcroft: Using abspath tends to make logging output cruftier than required, but otherwise should almost always be OK.

ska_conda/ska_builder.py
jeanconn: I'd prefer to run the tests and fix any tests that need fixing, instead of skipping the tests with "--no-test".

ska_conda/ska_builder.py
jeanconn: For verification of the behavior, I had just done a single test of this outside this code:

 - built Ska.Shell at 3.3.1
 - tried to build again with --skip-existing, but no build was done
 - checked out the repo at 3.3.2
 - tried to rebuild with --skip-existing and it built 3.3.2

Hopefully there are no gotchas.

jeanconn: I think this is not going to help conda to "know" when any meta packages should be rebuilt, so if we want that automated, we will need to figure out a mechanism.

ska_conda/ska_builder.py
jeanconn: Oh, and testing or not, I think we should check the status of the conda build command and do something with it (stop or save to report at the end).

taldcroft: As you know I'm not a big fan of build time testing, but we can discuss later when we have the core requirements all in place.

:+1: on checking the build command status.  Just call with `check=True, timeout=120` so it raises an exception if anything went wrong or it takes too long.

jeanconn: Maybe a test option for the build?  Mentioning because we haven't defined a process to make new recipes.  If we just plop them in the repo, the easiest way to determine they are complete and correct is to run the tests.  For example, by running the build tests, I just discovered that the maude recipe needs to have the requests package added as a runtime dependency/requirement.  Of course, I could run the build and build tests outside the ska_builder process, but I think that might be over-complicated.

jeanconn: Also, that timeout won't work for some builds, so I'm not sure if it would be better to just not define one for now.

jeanconn:
I think I'd like to merge this and do some other changes in separate/smaller PRs.  Were there any outstanding issues you really wanted to see addressed in this PR @taldcroft  ?

ska_conda/ska_builder.py
jeanconn: It looks to me that we need "--tags" to be safe if a tag has actually changed on origin.  For example, I retagged skare3 repo and I wasn't getting the associated new commit until adding this back again.  I know this shouldn't come up (you should re-use a tag), but...

Could also delete all local tags before the fetch, but that seems more problematic.

PR #7, opened 2018-06-30
Bulk changes to the recipes and build order.

### LINKED ISSUE #3: kadi package provides Chandra.Time
Looks like a copy-paste oops in `meta.yaml`.
```
(ska3) neptune$ tar tvf kadi-3.20.1-py36_0.tar.bz2 
...
-rw-rw-r--  0 jzuhone staff       1 May 22 14:37 lib/python3.6/site-packages/Chandra.Time-3.20.1-py3.6-macosx-10.9-x86_64.egg-info/not-zip-safe
...
```

### LINKED ISSUE #4: Django issues
Now the fun begins.

Django is a package that can't be allowed to float even in an uncontrolled bleeding edge Ska3.  Django 2.x definitely does not work.  I tried asking conda to downgrade to 1.6 but it wasn't happy:
```
(ska3) neptune$ conda install django=1.6
Solving environment: failed

UnsatisfiableError: The following specifications were found to be in conflict:
  - django=1.6
  - xija
Use "conda info <package>" to see the dependencies for each package.
```
Conda would accept django 1.8, so I tried that.  That runs tests with some failures, though I'm not sure if they are due to using 1.8 or due to my standalone mac environment.

So I force uninstalled django 1.8 and installed django 1.6.1 from the tarball.  That was worse:
```
In [1]: import kadi

In [2]: kadi.test()
=========================================================== test session starts ===========================================================
platform darwin -- Python 3.6.5, pytest-3.6.2, py-1.5.4, pluggy-0.6.0
rootdir: /Users/aldcroft/miniconda3/envs/ska3/lib/python3.6/site-packages, inifile:
plugins: remotedata-0.3.0, openfiles-0.2.0, doctestplus-0.1.2, arraydiff-0.2
collected 35 items / 1 errors                                                                                                             

================================================================= ERRORS ==================================================================
_______________________________________________ ERROR collecting kadi/tests/test_events.py ________________________________________________
kadi/tests/test_events.py:6: in <module>
    from .. import events
kadi/events/__init__.py:59: in <module>
    from .query import *
kadi/events/query.py:11: in <module>
    from . import models
kadi/events/models.py:9: in <module>
    from django.db import models
django/db/models/__init__.py:5: in <module>
    from django.db.models.query import Q
django/db/models/query.py:14: in <module>
    from django.db.models.fields import AutoField
django/db/models/fields/__init__.py:15: in <module>
    from django import forms
django/forms/__init__.py:8: in <module>
    from django.forms.fields import *
django/forms/fields.py:17: in <module>
    from django.forms.util import ErrorList, from_current_timezone, to_current_timezone
django/forms/util.py:4: in <module>
    from django.utils.html import format_html, format_html_join
django/utils/html.py:14: in <module>
    from .html_parser import HTMLParser, HTMLParseError
django/utils/html_parser.py:12: in <module>
    HTMLParseError = _html_parser.HTMLParseError
E   AttributeError: module 'html.parser' has no attribute 'HTMLParseError'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
========================================================= 1 error in 1.36 seconds =========================================================
```
Of course kadi events was working before in my mac standalone Ska3, but just yesterday I did some housecleaning and blew it away.

jeanconn:
Bulk changes to the recipes and build order.

Fixes #3 
Fixes #4


jeanconn:
I started with individual commits and then it just got silly.

jeanconn:
For a few of builds, I installed with pip instead of `python setup.py install` because it made the module tests work.  If we don't care about the tests, I'm not sure if we care about the install, but I suspect this has something to do with egg dirs or whatnot and we should probably check that the built packages match what is somewhat expected to traverse them with `testr`.

taldcroft:
I had assumed that using `pip install` per the spec in `python_modules.cfg` was the always the way to go.  But I don't know if there are any unexpected interactions with conda.

jeanconn:
And yes, for the few that failed I use the `pip install` per the python_modules.cfg spec.  I think we can change them all, we'd probably just need to respect `--egg`.  And yes, concern about interaction with the meta.yaml `preserve_egg_dirs` is why I didn't change this in bulk, though it should be easy to rebuild and retest if that's the way we end up going.

ska_conda/pkg_defs/ska-flight/meta.yaml
```diff
@@ -0,0 +1,38 @@
+package:
+  name: ska-flight
+  version: 0.1
+
+requirements:
+  # Packages required to run the package. These are the dependencies that
+  # will be installed automatically whenever the package is installed.
+  run:
+    - ska.shell ==3.3.3
```
taldcroft: Alphabetize for convenience in finding a package to update version.

ska_conda/pkg_defs/mica/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - testr
+  # Packages required to run the package. These are the dependencies that
+  # will be installed automatically whenever the package is installed.
+  run:
+    - python
+    - six
+    - setuptools
+    - matplotlib
+    - numpy
+    - pytest
```
taldcroft: In general seems this should be `testr`, which is directly imported in the package and then requires `pytest`.

ska_conda/pkg_defs/mica/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - django
```
taldcroft: ???

ska_conda/pkg_defs/mica/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  build:
+    - python
+    - setuptools
+    - testr
+  # Packages required to run the package. These are the dependencies that
+  # will be installed automatically whenever the package is installed.
+  run:
+    - python
+    - six
+    - setuptools
```
taldcroft: `setuptools` should not be an explicit runtime requirement.

ska_conda/pkg_defs/maude/meta.yaml
```diff
@@ -31,17 +30,19 @@ requirements:
   run:
     - python
     - setuptools
+    - requests
     - six
     - scipy
     - numpy
-    - testr
```
taldcroft: `testr` is required to do `import maude; maude.test()`.

ska_conda/pkg_defs/kadi/meta.yaml
```diff
@@ -29,7 +28,8 @@ requirements:
     - pytables
     - configobj
     - requests
-    - django
+    - tables3_api
+    - django <2.0
```
taldcroft: Nice way to encapsulate this locally!

ska_conda/build_order.txt
```diff
@@ -3,22 +3,29 @@ Ska.File
...
-#chandra_aca
-ska
\ No newline at end of file
+chandra_aca
+Ska.arc5gl
+Ska.astro
+mica
+cxotime
+annie
+ska-flight
```
taldcroft: The interface doc specifies `ska3-flight`.  I went back and forth about whether to try being consistent with `ska3` branding instead of `ska` to make this distinction clear to ourselves and end-users.  In a few years the `3` might seem superfluous, but for now when we say "Ska flight" that is really Ska2 flight, so probably worth being rigorous in here.

ska_conda/pkg_defs/ska-flight/meta.yaml
jeanconn: Good point.  We may end up doing this in other packages too, as largely they just accumulated dependencies as they were considered.

ska_conda/pkg_defs/mica/meta.yaml
jeanconn: For a bunch of packages, we have imports that are only used in certain circumstances.  For mica, django is needed for mica.web.  Obviously mica.web is useless unless we're running kadi webserver, in which case there is already a django.

I agree it is probably fine to remove django from here.

ska_conda/pkg_defs/mica/meta.yaml
jeanconn: Will check.

ska_conda/pkg_defs/maude/meta.yaml
jeanconn: I went back and forth on this and can obviously go back again.  It did seem to me that testr is more of a test dependence than a runtime dependence, and in a real ska runtime environment, there's always going to be a testr... but if we want to make sure that each ska package is good about also pulling in testr from the dependency tree.. that's fine too.

ska_conda/pkg_defs/mica/meta.yaml
jeanconn: It looks like the mica test packages import pytest directly so I think it is fine to have them both (testr and pytest) even if you end up with redundancy in the dependency tree.

ska_conda/pkg_defs/Chandra.Maneuver/meta.yaml
```diff
@@ -30,12 +31,15 @@ requirements:
...
     - setuptools
     - six
-    - testr
     - quaternion
     - chandra.time
     - ska.sun
 
 test:
+  requires:
+    - agasc
```
jeanconn: This is a case where we have some involved testing that seems to need agasc and Ska.engarchive.  I suppose it would be fine to just add those as runtime dependencies, but it seemed overkill (they don't really seem like run-time requirements for the package even if they are needed to complete all of its tests).

ska_conda/pkg_defs/maude/meta.yaml
taldcroft: If you are going to the trouble of enumerating Ska package dependencies at all, we should make them complete and not depend on being in a "real" Ska runtime.  Otherwise we could pretty much skip all the `meta.yaml` `run` requirements (except for build-time testing).

ska_conda/pkg_defs/Chandra.Maneuver/meta.yaml
taldcroft: Interesting web we are weaving.  Agreed this is fine to leave as is.

ska_conda/pkg_defs/maude/meta.yaml
jeanconn: OK, I'll move more of the test dependencies into the runtime area.

jeanconn:
I've got some more 'misc' type stuff and that setuptools tweak and such that can probably go in here , but if you are waiting on this to do something, I can just put them in a new PR.

PR #18, opened 2018-07-09
More recipe tweaks

jeanconn:
Recipe tweaks, including moving most of the build scripts into meta.yaml.

### DIFF
--- a/ska_conda/pkg_defs/Chandra.Maneuver/build.sh
+++ b/ska_conda/pkg_defs/Chandra.Maneuver/build.sh
@@ -1,2 +1,0 @@
-python setup.py install
-
--- a/ska_conda/pkg_defs/Chandra.Maneuver/meta.yaml
+++ b/ska_conda/pkg_defs/Chandra.Maneuver/meta.yaml
@@ -3,6 +3,7 @@ package:
   version:  {{ GIT_DESCRIBE_TAG }}
 
 build:
+  script: "pip install --no-deps --verbose --no-binary :all: --no-index --egg ."
   noarch: python
   script_env:
     - SKA_TOP_SRC_DIR
@@ -18,6 +19,7 @@ requirements:
   # Packages required to build the package. python and numpy must be
   # listed explicitly if they are required.
   build:
+    - pip <10.0
     - python
     - setuptools
     - numpy
@@ -29,7 +31,6 @@ requirements:
   # will be installed automatically whenever the package is installed.
   run:
     - python
-    - setuptools
     - six
     - quaternion
     - chandra.time
--- a/ska_conda/pkg_defs/Chandra.Time/build.sh
+++ b/ska_conda/pkg_defs/Chandra.Time/build.sh
@@ -1,1 +1,0 @@
-python setup.py install
--- a/ska_conda/pkg_defs/Chandra.Time/meta.yaml
+++ b/ska_conda/pkg_defs/Chandra.Time/meta.yaml
@@ -3,11 +3,10 @@ package:
   version:  {{ GIT_DESCRIBE_TAG }}
 
 build:
+  script: "pip install --no-deps --verbose --no-binary :all: --no-index --egg ."
   script_env:
     - SKA_TOP_SRC_DIR
   preserve_egg_dir: yes # (default no)
-  entry_points:
-    - entry = conda_version_test.manual_entry:main
 
 source:
   path: {{ SKA_TOP_SRC_DIR }}/Chandra.Time
@@ -20,7 +19,7 @@ requirements:
   # listed explicitly if they are required.
   build:
     - python
-    - setuptools
+    - pip <10.0
     - cython >=0.20.1
     - six
     - numpy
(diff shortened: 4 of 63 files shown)


ska_conda/pkg_defs/ska3-template/bin/flt_envs
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python -E
+#!PREFIXPYTHON -E
```
jeanconn: Could use /usr/bin/python here as well, I just figured there was less of a reason to do so if we are guaranteed to have a Python in our conda environment by the recipe requirements.  So the build.sh for this recipe just substitutes in the prefix.  At least that is the plan.

ska_conda/pkg_defs/ska3-template/bin/flt_envs
```diff
@@ -28,28 +27,18 @@ if SHELL not in ['csh', 'tcsh', 'bash', 'sh', 'ksh']:
...
-
 # Use the convention that flt_envs must be in $SKA/bin to derive SKA and
 # OVERRIDE any existing value.
 ska = dirname(dirname(abspath(__file__)))  # from $SKA/bin/flt_envs to $SKA
 
 # Set environment vars
 envs = dict(('SKA_' + suffix.upper(), os.path.join(ska, suffix))
         for suffix in ('arch', 'data', 'lib', 'share', 'bin', 'idl'))
-ska_arch_os = os.path.join(ska, 'arch', uname)
-envs['SKA_ARCH_OS'] = ska_arch_os
```
jeanconn: For the No arch install, do we have any reason to define SKA_ARCH_OS = SKA or is it best to get rid of it altogether as I've done here?

ska_conda/pkg_defs/ska3-template/bin/flt_envs
jeanconn: I rebased these changes away for now and put them in a new PR.  Will come back to this.

taldcroft:
What's the driver for moving the build scripts into `meta.yaml`?  I'm worried about eventual support for Windows.

jeanconn:
Sure.  For me the drivers were visibility and simplicity.  Basically, I missed an "--egg" and I didn't want to do that again.   I also thought I was doing a good job of moving toward simplicity by not trying to support platforms I can't test.

However, you are probably right that for ease of extensibility it will be easier to plop in a bld.bat (or whatever) than to edit the meta.yaml files with an os selector.  I'll clean this up and put them back.

taldcroft:
The plan would be for @jskrist to support Windows builds and do the testing.  I think that using (what I believe is) the standard recipe idiom makes things generally simpler.  For instance you could change all the build.sh to one of two flavors (egg/no-egg) with three commands (untested!):
```
echo ska_conda/pkgs_defs/*/build.sh | xargs -n 1 cp build-no-egg.sh
echo ska_conda/pkgs_defs/Chandra.*/build.sh | xargs -n 1 cp build-egg.sh
echo ska_conda/pkgs_defs/Ska.*/build.sh | xargs -n 1 cp build-egg.sh
```
Likewise the same with a build-no-egg.bat and build-egg.bad

jeanconn:
Yeah, that's fine.  And that's about what I did to do the egg vs no egg builds.
I got the impression from the docs that the standard recipe idiom had moved to just putting it in meta.yaml for a one-liner, but fine with separate build scripts for each platform instead of having selectors in the meta.

jskrist:
I'm happy to help, but that will be a bit down the road for me.  Separate scripts per OS are fine with me, but does add a bit of overhead (changing the build might require changing multiple files in pretty much the same way).

taldcroft:
@jeanconn - Are there more recently-updated docs than these tutorials?
  https://conda.io/docs/user-guide/tutorials/index.html
These seem to only talk about the three files for building (meta.yaml, build.{sh,bat}).

@jskrist - Yup, just looping you in to the conversation, no action required.

jeanconn:
Yeah, I have no idea if there are more recently-updated docs.  My approach has largely been to look around at example recipes and then see what the docs support as allowed behavior.  I suppose I may have been influenced by the conda-forge notes even if they aren't conda canon

https://conda-forge.org/docs/recipe.html#optional-bld-bat-and-or-build-sh

jeanconn:
Also, I'm not sure if I answered your question, because when I just stuck the build string in the meta.yaml, that was still using one of the three files for building.   And by selectors, I just meant you could use 

https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html?highlight=preprocessing#preprocess-selectors

per the meta.yaml convention and the script suggestion:

https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html?highlight=preprocessing#script

taldcroft:
Oooh, excellent resource!!  Seeing them explicitly recommending pip in this context is enlightening.  So it may well be that putting the one-liner in meta.yaml may work for Windows as well in most cases.

taldcroft:
I just checked out the Windows build process based on link that @mbaski supplied in email.  It looks like a simple `python setup.py install`, probably not caring that everything gets installed in an egg.

jeanconn:
Also yeah, the conda-forge folks seem to have good process and notes based on experience working around conda's quirks.

PR #19, opened 2018-07-10
One more pass through the recipes, plus work toward the metapackages

jeanconn:
One more pass through the recipes, plus work toward the metapackages .


taldcroft:
Sorry...  Part of the original driver for moving to pip was to allow installation of packages into an existing conda environment in a way that conda could manage and be aware of.  When we were talking about pip for  building packages, I had a vague feeling that it isn't quite right for that purpose.  It looks like there is a standard template they recommend that works for non-namespace packages:

https://conda.io/docs/user-guide/tutorials/build-pkgs.html#writing-the-build-script-files-build-sh-and-bld-bat

For namespace packages I think that just dropping the options (`--single-version-externally-managed --record=record.txt`) and using plain `$PYTHON setup.py install` should work, but clearly needs to be tested.

jeanconn:
OK.  I had run into the problem that testr wasn't working on at least one conda package in navigating the tree when installed via setup.py install, but worked if the conda package was made with pip install.  I'll find that error again if it persists and we can decide if we want to modify the package or the install process.

taldcroft:
Ok I didn't realize there was an actual problem somewhere.  If you haven't already done it we can leave this as an issue for cleanup or investigation later.

jeanconn:
Right, and I didn't look hard at the issue or document the error I saw previously when I had that text from you that gave me the impression that you thought pip was the way to go.  And yes, I figured later I'd switch all the packages to the recommended "setup.py" build string,  install into a clean environment, and compare ska_testr results.

jeanconn:
Status at this point is something like this:

```**************************************************************
***      Package                  Script            Status ***
*** ------------------ ---------------------------- ------ ***
***   Chandra.Maneuver                 test_unit.py   pass ***
***   Chandra.Maneuver           post_check_logs.py   pass ***
***       Chandra.Time                 test_unit.py   pass ***
***       Chandra.Time           post_check_logs.py   pass ***
*** Chandra.cmd_states                 test_unit.py   pass ***
*** Chandra.cmd_states           post_check_logs.py   pass ***
***         Quaternion                 test_unit.py   pass ***
***         Quaternion           post_check_logs.py   pass ***
***            Ska.DBI                 test_unit.py   pass ***
***            Ska.DBI           post_check_logs.py   pass ***
***          Ska.Numpy                 test_unit.py   pass ***
***          Ska.Numpy           post_check_logs.py   pass ***
***        Ska.ParseCM             test_unit_git.sh   pass ***
***        Ska.ParseCM           post_check_logs.py   pass ***
***          Ska.Shell                 test_unit.py   pass ***
***          Ska.Shell           post_check_logs.py   pass ***
***          Ska.Table             test_unit_git.sh   ---- ***
***          Ska.Table           post_check_logs.py   ---- ***
***     Ska.engarchive    test_make_archive_long.sh   ---- ***
***     Ska.engarchive                 test_unit.py   pass ***
***     Ska.engarchive           post_check_logs.py   pass ***
***            Ska.ftp                 test_unit.py   pass ***
***            Ska.ftp           post_check_logs.py   pass ***
***       Ska.quatutil             test_unit_git.sh   pass ***
***       Ska.quatutil           post_check_logs.py   pass ***
***            Ska.tdb                 test_unit.py   pass ***
***            Ska.tdb           post_check_logs.py   pass ***
***       acisfp_check              test_regress.sh   ---- ***
***       acisfp_check           post_check_logs.py   ---- ***
***       acisfp_check              post_regress.py   ---- ***
***              agasc                 test_unit.py   pass ***
***              agasc           post_check_logs.py   pass ***
***        chandra_aca                 test_unit.py   pass ***
***        chandra_aca           post_check_logs.py   pass ***
***            cxotime                 test_unit.py   pass ***
***            cxotime           post_check_logs.py   pass ***
***          dea_check              test_regress.sh   ---- ***
***          dea_check           post_check_logs.py   ---- ***
***          dea_check              post_regress.py   ---- ***
***          dpa_check              test_regress.sh   ---- ***
***          dpa_check           post_check_logs.py   ---- ***
***          dpa_check              post_regress.py   ---- ***
***               kadi         test_regress_long.sh   ---- ***
***               kadi                 test_unit.py   FAIL ***
***               kadi           post_check_logs.py   FAIL ***
***               kadi         post_regress_long.py   ---- ***
***              maude                 test_unit.py   pass ***
***              maude           post_check_logs.py   pass ***
***               mica                 test_unit.py   pass ***
***               mica           post_check_logs.py   pass ***
***   package_manifest              test_regress.sh   pass ***
***   package_manifest           post_check_logs.py   pass ***
***   package_manifest              post_regress.py   pass ***
***           parse_cm                 test_unit.py   pass ***
***           parse_cm           post_check_logs.py   pass ***
***             pyyaks                 test_unit.py   pass ***
***             pyyaks           post_check_logs.py   pass ***
***          starcheck              test_regress.sh   ---- ***
***          starcheck           post_check_logs.py   ---- ***
***          starcheck              post_regress.py   ---- ***
***          timelines test_unit_git_sybase_long.sh   ---- ***
***          timelines           post_check_logs.py   ---- ***
***               xija                 test_unit.py   pass ***
***               xija           post_check_logs.py   pass ***
**************************************************************
```
That's with a symlink to 'data' and using this test spec

```
(ska3) bash-4.1$ more test_spec_PY3 
# Excludes
- find_attitude
- dpa_check
- dea_check
- acisfp_check
- kadi*long*
- Ska.engarchive*long*
- timelines
- starcheck
- Ska.Table
```

Failures on kadi are due to the ftp tests.

jeanconn:
Oh and I was using something like this for version get_version_id for myself for now
```
#!/bin/sh
echo `conda list ska3-flight --export | tail -n 1`_`basename ${SKA}`_${HOSTNAME}
```

scripts/ska3_builder
```diff
@@ -10,20 +10,20 @@ if os.uname().sysname == "Darwin":
 
 parser = argparse.ArgumentParser(description="Build Ska Conda packages.")
 
-parser.add_argument("mode", type=str, help="The build mode. Either 'all' for all "
-                                           "packages, 'updated' for updated "
-                                           "packages, or a specific package name.")
-parser.add_argument("--build_root", default=".", type=str, 
+parser.add_argument("package", type=str, nargs="?",
+                    help="Package to build.  All we be built as needed if no package supplied")
```
taldcroft: `All will be built if no package supplied.`  ("As needed" is not needed).

scripts/ska3_builder
```diff
@@ -10,20 +10,20 @@ if os.uname().sysname == "Darwin":
...
+parser.add_argument("--tag", type=str,
+                    help="Optional tag, branch, commit for git checkout.  Only applies to single package build")
```
taldcroft: `Optional tag, branch, or commit to build for single package build (default is tag with most recent commit)`

scripts/ska3_builder
```diff
@@ -10,20 +10,20 @@ if os.uname().sysname == "Darwin":
...
+parser.add_argument("--build-root", default=".", type=str,
                     help="The path to the root directory for the Ska packages. "
```
taldcroft: `Path to root directory for output conda build packages.`

ska_conda/pkg_defs/ska3-template/bin/flt_envs
```diff
@@ -0,0 +1,79 @@
...
+    raise ValueError("shell '%s; not supported" % SHELL)
+
+
+# Use the convention that flt_envs must be in $SKA/bin to derive SKA and
+# OVERRIDE any existing value.
+ska = dirname(dirname(abspath(__file__)))  # from $SKA/bin/flt_envs to $SKA
+
+# Set environment vars
+envs = dict(('SKA_' + suffix.upper(), os.path.join(ska, suffix))
+        for suffix in ('arch', 'data', 'lib', 'share', 'bin', 'idl'))
```
taldcroft: `SKA_ARCH_OS` is definitely needed and should be the same as `SKA`.

ska_conda/ska_builder.py
```diff
@@ -17,16 +18,19 @@ def __init__(self, build_root='.', user='sot', git_repo_path='git@github.com:{us
...
         clone_path = os.path.join(self.src_dir, name)
         if not os.path.exists(clone_path):
             # Try ssh first to avoid needing passwords for the private repos
             # We could add these ssh strings to the meta.yaml for convenience
+            # The default ssh is not going to work if the package name doesn't match the
+            # top-level git repo name (cmd_states, eng_archive)
             try:
                 repo = git.Repo.clone_from(self.git_repo_path.format(user=self.user, name=name), clone_path)
                 assert not repo.bare
+                print("Clonied via default ssh git")
```
taldcroft: `Cloned`

scripts/ska3_builder
jeanconn: That's not strictly true if we use the "--skip-existing" flag to conda build (which I like).  I think it will only build if there is a new tag or if the recipe has changed, but  haven't explicitly checked the "recipe has changed" path.

ska_conda/pkg_defs/ska3-template/bin/flt_envs
jeanconn: Is it though?  Did pretty well on the pass through ska_testr tests without it and *not* setting it here might be a good way to flush out other implicit dependencies.

jeanconn: I'm also not sure if we need to be using that loop just above here that is apparently setting SKA_ARCH and SKA_IDL.

taldcroft: Right now the `acis_taco` cron job will definitely fail:
https://github.com/search?q=org%3Asot+ska_arch_os&type=Code

There are enough other references that I'd rather just leave this in for now and clean up later.

taldcroft: I originally was going to say remove `arch` and `idl` but have been trying (for your sake) to let go of minor nits that can be cleaned later.  I had a number of other comments that also went unsaid...  But we do still want SKA_DATA (and lib, share, bin).

jeanconn: That's fine.  I was just first shoot for "passes ska_testr" tests.  I think we should probably expand the tests to have more coverage.  

Also, starcheck is not PY3-working code yet, so there would be a bunch of changes to that project to work with this new env anyway.

scripts/ska3_builder
taldcroft: Sorry.  `All updated packages will be ...`.  The recipe changing is a corner case that doesn't need to be explicitly mentioned or tested now.

ska_conda/pkg_defs/ska3-template/bin/flt_envs
taldcroft: We are in the realm where this is basically close and by far the most efficient path for finding problems is to wrap it up and start using it.  This does not have to be perfect in version 0.1 (i.e. we can leave some non-functioning vars and cruft) and we don't have to think through every possible problem right now.

taldcroft:
> my thought was that we'd keep ska3-pinned for use with ska3-dev, but maybe that's just confusing. This will need more work due to the issue that we can't really pin conda in a generic way with ska3-pinned anyway, as it looks like you can't install conda into non-root envs

Not a problem.  What's here is fine, if we actually have a problem in the future we'll address it.

taldcroft:
> Not sure what to do with default ska data at this point.

This is indeed a slight issue, but OK as is.  See #22.

scripts/ska3_builder
jeanconn: OK!  "All updated packages will be built" sounds a bit circular to me (this is called the package update process in the docs,  but I'll run with that.

jeanconn:
Interesting to note that the new tags in Chandra.cmd_states, kadi, and xija broke my "clean build with ska3_builder and test ska3-flight" test because I didn't have the current/approved versions of those packages in my build area (and now just had the testing tagged builds).   I just supplied the flight tags for those to finish building and testing.

taldcroft:
I'm not totally pysched on e3ed543, but can accept this for now.  I don't think our process requires building packages at a non-tagged commit, but maybe you have a different idea.  In any case, the version should be like `3.4.2.dev3` not `3.4.2-dev3` to be PEP440 compliant. 

My view of process is now at https://github.com/sot/skare3/wiki/Ska3-runtime-environment-for-maintainers#package-update-testing-and-control-process.  This needs more detail and fine-tuning of course.

taldcroft:
> Interesting to note that the new tags in Chandra.cmd_states, kadi, and xija broke my "clean build with ska3_builder and test ska3-flight"

That is a fairly unusual situation, but good to see that there is a trivial workaround that we can apply.

jeanconn:
"I don't think our process requires building packages at a non-tagged commit, but maybe you have a different idea."  You are right on that score and this was definitely a belt-and-suspenders commit.   I was just kicking the tires and confirmed that if you supplied a no-tagged commit via sha you could shoot yourself in the foot.  I grant it would take some work.  If we need the protection we could also add code to only allow git checkout of tagged commits.

Also fine with reverting or not making the change in e3ed543 .

jeanconn:
Regarding `ska_version` my run_testr idea was to use something based on the ska3-flight version.

However, the next thing I did was to make a conda ska3 with updated xija, kadi, cmd_states packages to try that as a mini integration testing environment.  And of course to install dev versions of those packages in my env, I had to uninstall ska3-flight.

taldcroft:
Thanks, I'm happier reverting.  Remember this tool is not going to be run by non-experts, and in all likelihood only you or me.  And my intent is to have a detailed process description that we are following, so this doesn't really need to be bulletproof.

taldcroft:
> However, the next thing I did was to make a conda ska3 with updated xija, kadi, cmd_states packages to try that as a mini integration testing environment. And of course to install dev versions of those packages in my env, I had to uninstall ska3-flight.

I was thinking more along the lines of:
```
conda create -n test-ska ska3-flight
source activate test-ska
conda uninstall --force xija kadi cmd_states
# Then pip install xija, kadi and cmd_states from git repos
```
Note the possibility of installing directly from github using a particular tag or branch:
https://coderwall.com/p/-wbo5q/pip-install-a-specific-github-repo-tag-or-branch
```
pip install -e git://github.com/{ username }/{ reponame }.git@{ tag name }#egg={ desired egg name }
```

jeanconn:
OK.  I thought using '--force- might be bad form for this because you end up with an environment with explicitly unmet dependencies.

  If not clear, uninstalling ska3-flight didn't uninstall all the packages it brought with it, it just uninstalled the metapackage itself.

taldcroft:
> If not clear, uninstalling ska3-flight didn't uninstall all the packages

Ah yes, that wasn't clear.  But even so my experience / understanding is that uninstalling e.g. kadi will also uninstall anything that kadi depends on.  So I thought the `--force` is needed anyway, and yes, you get a temporarily "broken" environment where some packages just won't work until you manually re-install kadi from source.

jeanconn:
Right, but in my test environment I'd just uninstalled ska3-flight and upgraded kadi, cmd_states, xija.  Since we don't have package versions on the rest of the dependencies, that was fine and no dependencies were unment.

taldcroft:
What was the "upgrade" process?  Just pip install over the conda package?  If that works, then that would be fine.

jeanconn:
Sorry, I meant that... because my brand new ska3_builder test had,  "accidentally" made conda packages of the 3 dev packages in question, I just conda installed the test packages after uninstalling ska3-flight.  So not what you wanted to do for pip install testing.

ska3-flight.  So
```
(ska3) bash-4.1$ conda install chandra.cmd_states --channel  file:///fido.real/tmp/rebuild/builds/
...
The following packages will be UPDATED:
    chandra.cmd_states: 3.14-pyhbde51a3_0 file:///fido.real/tmp/rebuild/builds --> 3.14.1-pyhbde51a3_0 file:///fido.real/tmp/rebuild/builds
```

jeanconn:
It looks like "pip install over the conda package" does not work.  I ended up with two kadi packages according to "conda list", which seems not maintainable.  So for pip installs of dev packages, your method of force uninstalling the conda packages and then pip installing seems exactly right.

taldcroft:
I updated the process Wiki accordingly.

taldcroft:
About ska_version for ska_testr with dev package(s), I see the issue there and don't have a great solution.  

One possibility is somehow in the build process getting the version of ska3-flight into an installed file e.g. $PREFIX/SKA_VERSION.  (This might make ska3-flight not really a meta-package if done directly there).  Then `ska_version` just echos what is in that file.  For testing the procedure would specify to edit the SKA_VERSION file to add some tag or whatever prior to running ska_testr.  I.e. if you think the new ska3-flight will be versioned as 2019.06.22 then use 2019.06.22rc1 in SKA_VERSION.

jeanconn:
The only problem with that plan to use ska3-flight to make a file, is that if you uninstall ska3-flight the version file would go away.  If you follow your suggested convention to just force remove other packages as needed packages while leaving ska3-flight in place, I guess that would be fine.

taldcroft:
BTW, I don't see the ska_version stuff as a lien on this PR.  As long as ska_testr is getting through to the end with *some* version, whatever it might be, that is good enough for here.

taldcroft:
@jeanconn - is this ready to merge?  I.e. no major problems as far as you know?

jeanconn:
Given that we have no dependence on master yet and I think we've discussed all the issues in this PR well in the context of this PR, yes, fine to merge.  We can handle other issues in other PRs.  

I have just been re-tagging along this branch with the temporary tag "0.1.0" as part of the build/test process, but I don't know if we want a more official tag at this point and then have other changes be as PRs on master and new bumps to the release/tag.

taldcroft:
:tada:  Fantastic!

jeanconn:
OK.  I just did a clean build on presto (centos 5) of our packages at current versions and updated the linux-64 and noarch channels with those builds.  I'll rebuild build on osx soon.  I moved the old stuff in the ska3-conda channel to a subdirectory; but from here forward I think we'll just be updating as we actually update packages.  Also, right now the metapackage all have dummy versions, which we'll want to replace with the date style versions you've suggested.

Also, I rebuilt on centos 5 mostly for the practice.  Build system shouldn't make a difference because we shouldnt' be doing much against the system libraries, especially for the noarch packages.  This may not be as true as I go back to finishing up the Perl pacakges.

PR #28, opened 2018-07-13
Add empty string as option value for sed -i

jeanconn:
Add empty string as option value for sed -i .  This fixes ska3-template creation on osx .

### DIFF
--- a/ska_conda/pkg_defs/ska3-template/build.sh
+++ b/ska_conda/pkg_defs/ska3-template/build.sh
@@ -4,5 +4,5 @@ for file in flt_envs ska_envs.sh ska_envs.csh
    do
     chmod +x ${RECIPE_DIR}/bin/${file}
     cp -a ${RECIPE_DIR}/bin/${file} ${PREFIX}/bin
-    sed -i "s|%{PREFIX}%|${PREFIX}|" ${PREFIX}/bin/${file}
+    sed -i ""  "s|%{PREFIX}%|${PREFIX}|" ${PREFIX}/bin/${file}
    done


taldcroft:
Got it.  BTW, I think that auto-closing issues requires putting it in the PR description, but in any case it is best practice to have it there for reference on creation of the PR.

taldcroft:
Now reading your comment, I think maybe about unapproving this and going to the simple solution.  We really don't have to build in contraptions to handle the extremely remote possibility of the conda prefix changing in the future.

ska_conda/pkg_defs/ska3-template/build.sh
```diff
@@ -4,5 +4,10 @@ for file in flt_envs ska_envs.sh ska_envs.csh
    do
     chmod +x ${RECIPE_DIR}/bin/${file}
     cp -a ${RECIPE_DIR}/bin/${file} ${PREFIX}/bin
-    sed -i ""  "s|%{PREFIX}%|${PREFIX}|" ${PREFIX}/bin/${file}
+    # Replace %{PREFIX}% with the conda build prefix.  It looks like it would
+    # be simpler to just make these three files with the standard conda prefix_placeholder
+    # hard-coded in the template files but this sed operation should be robust if the
```
jeanconn: By prefix_placeholder here, I mean that it does/could work to just hard code "/opt/anaconda1anaconda2anaconda3" as the prefix in the three files, as conda recognizes that prefix placeholder as a thing that should be replaced when actually installing the package.  But doing this does not seem to be advertised in the conda docs and I don't know if prefix_placeholder could change.

jeanconn:
There seem to be a lot more recipes that use sed to do something like what we were doing, but I'm fine with trying this out and if we get into a pickle we can try something else.

jeanconn:
I think I need to go back to sed.  Ugh.  Looks like the script in a non-root env isn't getting the install and rewrite correctly.

(ska3) bash-3.2$ which ska_envs.sh
/tmp/fullconda/envs/ska3/bin/ska_envs.sh
(ska3) bash-3.2$ more /tmp/fullconda/envs/ska3/bin/ska_envs.sh
eval `/tmp/fullconda/bin/flt_envs -shell sh -ska`

taldcroft:
OK, looks like we are officially in a pickle.

PR #32, opened 2018-07-16
Remove some packages and explicit version deps in ska3-core

jeanconn:
This removes dbus, gst-plugins-base, gstreamer, mpich2, and libxcb from
the ska3-core list, as those packages are only available on Linux in
the defaults channel.  Packages html5lib, zeromq, and libsodium have
been put in as unversioned requirements, as they have different versions
on Linux and OSX (we could probably use os selectors but may not care).

### DIFF
--- a/ska_conda/pkg_defs/ska3-core/meta.yaml
+++ b/ska_conda/pkg_defs/ska3-core/meta.yaml
@@ -9,6 +9,9 @@ requirements:
   # Packages required to run the package. These are the dependencies that
   # will be installed automatically whenever the package is installed.
   run:
+   - html5lib
+   - libsodium
+   - zeromq
    - alabaster ==0.7.10
    - asn1crypto ==0.22.0
    - astroid ==1.5.3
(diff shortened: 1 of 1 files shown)


taldcroft:
With the "arch versions", what would that end up looking like?  `ska3-flight-mac` and `ska3-flight-linux`, `ska3-core-mac`, `ska3-core-linux`?  Or something nicer or completely different?

I do still have this niggling concern, not about different package versions for OS's but unexpected / untested package version updates when installing a fresh Ska3 for whatever reason.  In practice I agree it is very unlikely that the unversioned packages here would actually cause a problem.  So I would prefer this over arch-specific metapackages (if they look anything like what I imagine).

jeanconn:
I think the arch-specific versions would look like

https://github.com/sot/skare3/pull/33

where we'd have one ska3-core recipe, but it would have OS-specific entries for a few elements using the preprocess selectors and you'd then build/update the recipe as an arch package not a "noarch" package.

jeanconn:
And right, I wasn't sure that installing from a set of defined packages was enough (installing from a ska3-core metapackage) if the metapackage install process can bring in untested dependencies or updates.  We could either experiment with conda install flags (--no-deps?) or have a separate task that confirms that the full package list is limited to exactly what is expected.

taldcroft:
@jeanconn - I think this can be closed.  Using #33 should also naturally result in the minimal core package set to support `ska3-flight` packages.

PR #33, opened 2018-07-17
ska3-core with arch-specific packages

### LINKED ISSUE #24: Move metapackages to noarch
As a follow-on to #19 and #16, plan to move these to noarch.  Even if the process for building is relatively quick in time, it still roughly doubles the effort to make an update (requiring logging into a separate build machine and rsyncing and blah blah).

### LINKED ISSUE #36: Confirm that ska3-core is the minimal set of packages
But consider if current packages might be there for other reasons besides supporting ska3-flight packages?

In general, fixing (adding) back packages seems a reasonable approach if we identify dependencies.

### LINKED ISSUE #42: Need nb_conda package for Jupyter notebook to work within env
If ska3 is installed as an env instead of the root, then jupyter notebook ends up using the root Python for the kernel and nothing works.  See https://github.com/jupyter/notebook/issues/3311#issuecomment-363516924 and subsequent comments for the resolution.  Installing it did the trick for me.

This brings in many more packages, and praying that it all works with our pinning.

jeanconn:
Configure the ska3-core metapackage recipe to have arch-specific dependencies as needed.

Closes #36 
Closes #24 
Closes #42

### DIFF
--- a/ska_conda/pkg_defs/ska3-core/meta.yaml
+++ b/ska_conda/pkg_defs/ska3-core/meta.yaml
@@ -2,9 +2,6 @@ package:
   name: ska3-core
   version: 2018.07.16
 
-build:
-  noarch: generic
-
 requirements:
   # Packages required to run the package. These are the dependencies that
   # will be installed automatically whenever the package is installed.
@@ -25,7 +22,7 @@ requirements:
    - cycler ==0.10.0
    - cython ==0.26
    - db ==5.3.28
-   - dbus ==1.10.20
+   - dbus ==1.10.20 [linux]
    - decorator ==4.1.2
    - django ==1.11.3
    - docutils ==0.14
@@ -35,11 +32,12 @@ requirements:
    - freetype ==2.5.5
    - git ==2.11.1
    - glib ==2.50.2
-   - gst-plugins-base ==1.8.0
-   - gstreamer ==1.8.0
+   - gst-plugins-base ==1.8.0 [linux]
+   - gstreamer ==1.8.0 [linux]
    - h5py ==2.7.0
    - hdf5 ==1.8.17
-   - html5lib ==0.9999999
+   - html5lib ==0.9999999 [linux]
+   - html5lib ==0.9999 [osx]
    - icu ==54.1
    - idna ==2.2
    - imagesize ==0.7.1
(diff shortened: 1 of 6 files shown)


ska_conda/pkg_defs/ska3-core/meta.yaml
```diff
@@ -35,11 +32,12 @@ requirements:
...
    - git ==2.11.1
    - glib ==2.50.2
-   - gst-plugins-base ==1.8.0
-   - gstreamer ==1.8.0
+   - gst-plugins-base ==1.8.0 [linux]
+   - gstreamer ==1.8.0 [linux]
    - h5py ==2.7.0
    - hdf5 ==1.8.17
-   - html5lib ==0.9999999
+   - html5lib ==0.9999999 [linux]
```
taldcroft: The `html5lib` folks need to commit to getting 1.0 out!!

jeanconn: Hah :smile:

jeanconn:
So the ska3-core updates in 0fb03f3 are what I get if I do a 
```
conda install ska3-dev ska3-pinned
```
on new environments on both linux and osx machines, and then run 
```
conda list --json > {os}.json
```
and then run the little script with those json files as input
```
python combine_arch.py --linux linux.json --osx osx.json --out package_list.txt
```
and then insert the package list into ska3-core.  Still testing the environment you get from that.  The little script uses the lowest common version if the package exists on both platforms, but does not use conda tools to confirm that package actually exists or is compatible with the other packages.  I'm just hoping.

taldcroft:
> I'm assuming we want to try to hold back packages on osx to somewhat match the linux versions (which are being held back by CentOS5) but I really don't know.

That seems like a reasonable starting point.  We'll see if problems come up in practice.

It's disturbing that a change in the conda bugfix version results in completely different package versions / builds.  But maybe 4.3.21 is the last one that doesn't go for the hashed-version builds?

jeanconn:
"But maybe 4.3.21 is the last one that doesn't go for the hashed-version build?"  I think that 4.3.25 is probably the last one that doesn't use the newer file areas that have the hash builds

https://github.com/conda/conda/blob/master/CHANGELOG.md#4326-2017-09-15

 But when I was looking for CentOS5 compatibility it looked like 4.3.21 was the last one actually built *on* the old os.  So I think it is a reasonable compromise for now. .

jeanconn:
Do you want changes to the combine_arch script or process here @taldcroft  or is this OK for this PR and we'll do more changes as needed in a new PR after testing?

taldcroft:
This is ok.  Approved.

PR #38, opened 2018-07-24
Simplify git cloning

### LINKED ISSUE #8: Move to using git ssh URLs in the recipes

jeanconn:
This PR removes the logic that tries to git clone packages using ssh by default, and instead just explicitly uses whatever string is after "home:" in the meta.yaml.  For private packages, that is now set to the github ssh string.  For the rest, it is set to the https url.

Closes #8 (via explicit compromise on https and ssh urls)

### DIFF
--- a/ska_conda/ska_builder.py
+++ b/ska_conda/ska_builder.py
@@ -24,7 +24,7 @@ class SkaBuilder(object):
             metayml = os.path.join(pkg_defs_path, name, "meta.yaml")
             # It isn't clean yaml at this point, so just extract the string we want after "home:"
             meta = open(metayml).read()
-            url = re.search("home:\s?(\S+)", meta).group(1)
+            url = re.search("home:\s*(\S+)", meta).group(1)
             repo = git.Repo.clone_from(url, clone_path)
             print("Cloned from url {}".format(url))
         else:
--- a/ska_conda/ska_builder.py
+++ b/ska_conda/ska_builder.py
@@ -70,11 +70,20 @@ class SkaBuilder(object):
         self._build_package(name)
 
     def build_list_packages(self):
+        failures = []
         with open(build_list, "r") as f:
             for line in f.readlines():
                 pkg_name = line.strip()
                 if not pkg_name.startswith("#"):
-                    self.build_one_package(pkg_name)
+                    try:
+                        self.build_one_package(pkg_name)
+                    # Just try to build the rest if there's a failure, but record the name
+                    except:
+                        failures.append(pkg_name)
+                        continue
+        if len(failures):
+            raise ValueError("Packages {} failed".format(",".join(failures)))
+
 
     def build_all_packages(self):
         self.build_list_packages()
--- a/ska_conda/ska_builder.py
+++ b/ska_conda/ska_builder.py
@@ -1,7 +1,7 @@
 import os
 import subprocess
 import git
-import yaml
+import re
 
 ska_conda_path = os.path.abspath(os.path.dirname(__file__))
 pkg_defs_path = os.path.join(ska_conda_path, "pkg_defs")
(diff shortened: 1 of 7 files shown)


ska_conda/ska_builder.py
```diff
@@ -10,43 +10,28 @@
...
-                    while not requires:
-                        line = f.readline().strip()
-                        if line.startswith("path:"):
-                            requires = True
-                    data = yaml.load(f)
-                    url = data['about']['home']
-                repo = git.Repo.clone_from(url, clone_path)
-                print("Cloned from url {}".format(url))
+            metayml = os.path.join(pkg_defs_path, name, "meta.yaml")
+            # It isn't clean yaml at this point, so just extract the string we want after "home:"
```
jeanconn: It looks like the previous `requires/path` loop was dealing with the fact that the meta.yaml isn't compliant YAML at this point at the `path` key.  I just went with a regex here to grab the string after `home:` instead of trying to use the YAML parser.  I think that is fair for our application.

ska_conda/ska_builder.py
```diff
@@ -85,11 +70,20 @@ def build_one_package(self, name, tag=None):
...
                 if not pkg_name.startswith("#"):
-                    self.build_one_package(pkg_name)
+                    try:
+                        self.build_one_package(pkg_name)
+                    # Just try to build the rest if there's a failure, but record the name
+                    except:
+                        failures.append(pkg_name)
+                        continue
+        if len(failures):
+            raise ValueError("Packages {} failed".format(",".join(failures)))
```
jeanconn: This isn't strictly required, but I thought it might be helpful to let the build process continue through the rest of the packages even if you have an issue with one package (an issue such as forgetting your ssh key password) and then printing out the failures and exiting with error status at the end.

ska_conda/ska_builder.py
taldcroft: I suspect it could be made into compliant YAML by enclosing the `{{ .. }}/..` stuff in double quotes, but I agree with your point to just get the job done here.

ska_conda/ska_builder.py
```diff
@@ -10,43 +10,28 @@
...
-                        if line.startswith("path:"):
-                            requires = True
-                    data = yaml.load(f)
-                    url = data['about']['home']
-                repo = git.Repo.clone_from(url, clone_path)
-                print("Cloned from url {}".format(url))
+            metayml = os.path.join(pkg_defs_path, name, "meta.yaml")
+            # It isn't clean yaml at this point, so just extract the string we want after "home:"
+            meta = open(metayml).read()
+            url = re.search("home:\s?(\S+)", meta).group(1)
```
taldcroft: Should be `\s*`.  Multiple spaces are fine.  I can't remember if zero spaces is OK for YAML, but for this purpose yes.

ska_conda/ska_builder.py
taldcroft: I sort of think that if something fails, there is really a problem that needs to be fixed before moving on (and maybe you can into dependency hell where every build starts failing because of the missing failed build).  Since it doesn't re-build everything every time (right?), it seems pretty cheap to restart.  The "forgetting ssh key" seems a not-credible failure.

ska_conda/ska_builder.py
```diff
@@ -10,43 +10,28 @@
...
         os.environ["SKA_TOP_SRC_DIR"] = self.src_dir
 
     def _clone_repo(self, name, tag=None):
         if name in no_source_pkgs:
             return
-        print("Cloning source %s." % name)
+        print("Cloning or updating source source %s." % name)
         clone_path = os.path.join(self.src_dir, name)
         if not os.path.exists(clone_path):
-            # Try ssh first to avoid needing passwords for the private repos
```
taldcroft: Awesome!

jeanconn: Regarding continuing, my thought was that there could be other reasons for failure that you might not care about if all you really wanted to do was update a couple of packages at the end of the list.  Or if you were a non-privileged user who'd just been handed this code base (JWST?) and this logic lets you really just see how everything works without stopping at parse_cm.  You could get in to dependency hell in a fresh build with no access to the currently build packages in ska3-conda, but the only fallout of this change in that case would just be that it would take a lot longer to finish failing out with the rest of the packages.

jeanconn: Also, "forgetting ssh key password" was a creditable failure in my testing.

taldcroft: OK, well then put in a text input to ask the user to continue ala `install.py`.

jeanconn: OK.

jeanconn:
Is this good to go @taldcroft  ?

PR #41, opened 2018-07-25
Add back use of sed for template rewrite

jeanconn:
Use sed to rewrite templates with the prefix of our installation.

This replaces the technique in #28


jeanconn:
Does this need anything explicit for testing besides just looking at the installed template files in both a root and non-root env and making sure that 

1. all the substitutions are right (and localized to the env)
2. flt_envs still has a "-E" in the shebang

?

PR #43, opened 2018-07-26
Add local astropy

jeanconn:
Bundle a recipe for astropy (and some test dependencies) in skare3 to give us an astropy 3.0.3 that will still run on our CentOS5-compatible Linux env.

### DIFF
--- a/ska_conda/pkg_defs/astropy/meta.yaml
+++ b/ska_conda/pkg_defs/astropy/meta.yaml
@@ -28,13 +28,16 @@ requirements:
   build:
     - python
     - setuptools
-    - numpy 1.12.*  # [py36]
-
+    - numpy ==1.12.1
+    - llvmlite ==0.18.0
+    - gcc ==4.8.5 [linux]
+    - libgcc ==4.8.5 [linux]
   run:
     - python
-    - numpy >=1.11  # [py36]
+    - llvmlite ==0.18.0
+    - numpy ==1.12.1
     - pytest-astropy
-
+    - libgcc ==4.8.5 [linux]
 test:
   commands:
     - fits2bitmap --help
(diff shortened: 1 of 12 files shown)


ska_conda/ska_builder.py
```diff
@@ -6,7 +6,9 @@
 ska_conda_path = os.path.abspath(os.path.dirname(__file__))
 pkg_defs_path = os.path.join(ska_conda_path, "pkg_defs")
 build_list = os.path.join(ska_conda_path, "build_order.txt")
-no_source_pkgs = ['ska3-flight', 'ska3-core', 'ska3-dev', 'ska3-pinned', 'ska3-template']
+no_source_pkgs = ['ska3-flight', 'ska3-core', 'ska3-dev', 'ska3-pinned', 'ska3-template',
```
jeanconn: Source probably rename 'no_source_pkgs' to something that makes it more clear that the packages in this list just don't use the local convention we're using that about: home: has the url for a git repository that we build a custom package from using the git tags.

ska_conda/pkg_defs/astropy/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - wcslint = astropy.wcs.wcslint:main
+  number: 1
+
+requirements:
+
+  build:
+    - python
+    - setuptools
+    - numpy ==1.12.1
+    - llvmlite ==0.18.0
```
jeanconn: I changed these to be very specific to our install.  I think this works out so that our recipe/astropy build will be naturally deprecated when we no longer have these pins.

ska_conda/pkg_defs/astropy/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - gcc ==4.8.5 [linux]
```
jeanconn: Also, didn't build on CentOS5 with gcc 4.1.2 .  This suggests to me that we should add a gcc/libgcc on our few other linux-64 packages.

Issue #45, opened 2018-07-26
Add ipywidgets 7.3.0 for Matt


jeanconn:

taldcroft:
Let's see if the work-around I proposed is sufficient for Matt.

jeanconn:
Closed by #46

PR #46, opened 2018-07-27
Make custom widgetsnbextension and ipywidgets

jeanconn:
Make custom widgetsnbextension and ipywidgets

These recipes (like astropy) are borrowed from conda-forge.


jeanconn:
These changes just tested on Linux and before Tom's last comment on #45

PR #47, opened 2018-07-27
Update metapackages for recent changes

jeanconn:
Update metapackages for recent changes

This brings ska3-flight a bit ahead of HEAD flight


jeanconn:
We'll probably cut at least one more ska3-flight/ska3-core depending on what we decide to do with ipywidgets, but things should be relatively correct and stable here to do some more testing.

PR #48, opened 2018-07-30
Update ska_version to include a sha of the other pkgs

jeanconn:
Update ska_version to include a sha of the other pkgs.  Looks like:

core-2018.07.16:flight-2018.07.16:pinned-2018.07.16:template-0.1:pkgs_sha-f272ef6

### DIFF
--- a/ska_conda/pkg_defs/ska3-template/bin/ska_version
+++ b/ska_conda/pkg_defs/ska3-template/bin/ska_version
@@ -1,19 +1,33 @@
 #!%{PREFIX}%/bin/python
 """
-Create and return an environment version string based on the ska3-* conda
-packages in an environment.
+Create and return an environment version string based on the packages
+in an environment.  The ska3-* package versions are called out explicitly,
+followed by a sha based on the versions of the other packages.
 
 Example:
 
-core-2018.07.16:flight-2018.07.16:pinned-2018.07.16:template-0.1
+core-2018.07.16:flight-2018.07.16:pinned-2018.07.16:template-0.1:pkgs_sha-f272ef6
 
 """
 
 import subprocess
 import json
+import hashlib
 # Conda list output appears already sorted by package name
 pkgs_json = subprocess.check_output(["conda", "list", "--json"])
 pkgs = json.loads(pkgs_json)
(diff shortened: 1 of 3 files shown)


ska_conda/pkg_defs/ska3-template/bin/ska_version
```diff
@@ -1,19 +1,33 @@
...
              for pkg in pkgs if pkg['name'].startswith('ska3-')]
+# Get a list of all the other conda pkgs and their versions (not builds or channels)
+all_pkgs = [f"{pkg['name']}-{pkg['version']}"
+            for pkg in pkgs if not pkg['name'].startswith('ska3-')]]
+# Make this list into a bytestring for hashlib
+all_pkgs = " ".join(all_pkgs).encode('utf-8')
+# Do a separate 'conda list' just to get the pip-installed pkgs
+conda_list = subprocess.check_output(["conda", "list"]).splitlines()
+for line in conda_list:
+    if b'<pip>' in line:
```
jeanconn: The check_output docs note that we can use "universal_newlines" to apparently get a string back instead of bytes https://docs.python.org/3.6/library/subprocess.html#subprocess.check_output

jeanconn: But as far as I can tell, hashlib needs bytes anyway, so I don't know if it matters.

taldcroft: Stylistically it's better to not change `all_pkgs` from a list to bytestring midstream, and incrementally appending strings is just something avoid.  I like the `encoding` argument here, which does all the string processing idiomatically and then uses a bytestring only when required.
```
pkgs = subprocess.check_output(["conda", "list"], encoding='utf-8').splitlines()
all_pkgs.extend(pkg for pkg in pkgs if '<pip>' in pkg)
sha_tag = hashlib.sha1(''.join(all_pkgs).encode('utf-8')).hexdigest()[0:7])
ska3_pkgs.append(f'sha-{sha_tag}')
```

taldcroft:
Unicode sandwich.  I found these slides to be very useful diving into Py3 and unicode:

https://nedbatchelder.com/text/unipain/unipain.html#35

PR #56, opened 2018-07-31
Add acis checkers

jeanconn:
Add acis checkers to skare3


taldcroft:
Looks like acis_taco is missing?  I assume that #57 doesn't prevent installing acis_taco for use by the FP model.

PR #58, opened 2018-07-31
Add ska_sync

jeanconn:
Add ska_sync


jeanconn:
Still need to bump versions in the metapackages.

ska_conda/pkg_defs/ska3-flight/meta.yaml
```diff
@@ -36,6 +36,7 @@ requirements:
     - ska.matplotlib ==3.11.2
     - ska.parsecm ==3.3.1
     - ska_path ==3.1
+    - ska_sync ==0.1.1
```
jeanconn: This has git tag 0.1.1 and version 0.1 in the code (as far as I can tell) which should probably get cleaned up at some point.

jeanconn:
Eh, I suppose I can just rebuild at the new metapackage versions from #56

PR #60, opened 2018-08-02
Make ska_version output shorter by default

taldcroft:
The dog-fooding begins...

The `:`-separated version actually breaks the Ska.engarchive long test because `:` in the path is interpreted as a path separator.  So using colon was just not a good choice.  But more to the point, I realized that in practice `ska3-flight` has embedded in it the other package versions.  Unless you purposely do crazy things with forcing different metapackages, the version of `ska3-flight` plus the SHA1 of everything is fine.

This PR makes that short form the default, but does add a `--full` option to get the full list consistent (mostly, see next paragraph) with the current output.

The other thing this changes is having *all* packages factor into the SHA hash.  This was mostly driven by having a simple code base that produces the same SHA for both the short and full forms, but also makes sense to me.  The SHA should be different if *any* package version is different.

### DIFF
--- a/ska_conda/pkg_defs/ska3-flight/meta.yaml
+++ b/ska_conda/pkg_defs/ska3-flight/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: ska3-flight
-  version: 2018.08.01
+  version: 2018.08.02
 
 build:
   noarch: generic
@@ -10,7 +10,7 @@ requirements:
   # will be installed automatically whenever the package is installed.
   run:
     - ska3-core ==2018.07.27.1
-    - ska3-template ==2018.07.30
+    - ska3-template ==2018.08.02
     - agasc ==3.5.2
     - annie ==0.5
     - acisfp_check ==v2.4.0
--- a/ska_conda/pkg_defs/ska3-template/bin/ska_version
+++ b/ska_conda/pkg_defs/ska3-template/bin/ska_version
@@ -6,24 +6,44 @@ followed by a sha based on the versions of the other packages.
 
 Example:
 
+$ ska_version
+2018.07.16-f272ef6
+
+$ ska_version --full
 core-2018.07.16:flight-2018.07.16:pinned-2018.07.16:template-0.1:sha-f272ef6
 
 """
-
 import subprocess
 import json
 import hashlib
+import argparse
+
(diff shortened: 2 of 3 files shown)


jeanconn:
Thanks Tom.   Regarding "The SHA should be different if any package version is different." sure, that makes sense.  If you've pulled the ska3- packages from the full version you are now capturing their versions in the SHA.  

The only bonus that we lose is that under the old system you could make a ska3-dev environment, have a sha that made sense on the non-ska3 packages, and then when you "promoted" that combination of packages the ska3-flight you'd still have the same non ska3- sha (you'd be able to easily see that the environment you originally tested in was equivalent).   I don't know if that will matter if we can just test ska3-flight without installing to ska3/flight .

taldcroft:
Yes, that's a fair point.  So I looked again and it was an easy fix in 2292679.

jeanconn:
Looks good.  I like the --full option, in that we can still use it in `ska_testr`, but not have applications (Ska.engarchive, starcheck regression tests) get too hung up on the long form string.

taldcroft:
Well, *you* can use it in ska_testr.  :smile:  I decided right away that the long form was just too long and that's what started me thinking about whether it is all really necessary.  Reduce reduce reduce.

BTW, do you think the maintainer instructions are in shape for me to try building and installing this update?

jeanconn:
> BTW, do you think the maintainer instructions are in shape for me to try building and installing this update?

Iffy.  That's a couple of questions:

1) do you want to set up an environment to build/install or do you want to use the current build environment in as aca on unagi in /disk1/telmon/conda (not finalized or in docs)?

2) do you want to use that environment to build a new version of ska3_builder to then build the other new packages out of pkg data, or do you just want to build the two new packages out of the git repo (process still not settled)?

taldcroft:
Sounds like a no.  I don't want to have to think, instead just go to a page and follow the instructions carefully.

jeanconn:
Though with a couple of these things, I think I'll still need you to run through more than one process to see which one you like better.

taldcroft:
Whichever is simpler and shorter.  I *think* that means below, but I'm not sure:

1.  Maintain a persistent build env like /disk1/..
2. Build out of git.

jeanconn:
I installed the updated template and flight packages to /proj/sot/ska3/test for continued testing.

taldcroft:
Ta.

PR #61, opened 2018-08-08
Add lxml per mta request

jeanconn:
Add lxml per mta request

This also adds the other dependencies as required: 
libiconv and libxml2 added to osx
libxslt and lxml added for both os

### DIFF
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: ska3-core
-  version: 2018.08.10
+  version: 2018.08.13
 
 requirements:
   # Packages required to run the package. These are the dependencies that
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -93,8 +93,10 @@ requirements:
    - libtiff ==4.0.6
    - libxcb ==1.12 # [linux]
    - libxml2 ==2.9.4 # [linux]
+   - libxslt ==1.1.29
    - line_profiler ==2.0
    - llvmlite ==0.18.0
+   - lxml ==3.8.0
    - markupsafe ==1.0
    - matplotlib ==2.0.2
    - mistune ==0.7.4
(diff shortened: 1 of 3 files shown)


taldcroft:
Did you try it out?  The use case is with bs4, so you can see discussion and examples below here:

https://www.crummy.com/software/BeautifulSoup/bs4/doc/#installing-a-parser

I think you just need to parse an HTML doc explicitly specifying the `lxml` parser, and that is sufficient.

jeanconn:
No, I'd just confirmed it imported without blowing up.  If you think we need to test added 'default' modules (more than numpy scipy) will do.

taldcroft:
The scope of my testing request was narrow and just for lxml.

jeanconn:
I don't actually see much in the way of examples that work as tests on the page you sent.

```
In [3]: xml_soup = BeautifulSoup('<p class="body strikeout"></p>', 'xml')
   ...: xml_soup.p['class']
   ...: 
Out[3]: 'body strikeout'

In [4]: xml_soup = BeautifulSoup('<p class="body strikeout"></p>', 'lxml')
   ...: xml_soup.p['class']
   ...: 
Out[4]: ['body', 'strikeout']

In [14]: from bs4.diagnose import diagnose

In [15]: with open("/tmp/bad.html") as fp:
    ...:     data = fp.read()
    ...: diagnose(data)
    ...: 
    ...: 
Diagnostic running on Beautiful Soup 4.6.0
Python version 3.6.2 |Continuum Analytics, Inc.| (default, Jul 20 2017, 13:51:32) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]
Found lxml version 3.8.0.0
Found html5lib version 0.999

```

ska_conda/pkg_defs/ska3-core/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: ska3-core
-  version: 2018.07.27.1
+  version: 2018.08.08
```
jeanconn: This will probably need another version bump by the time we get to it.

jeanconn:
And is the scope of testing complete or did you have more in mind @taldcroft  ?

PR #64, opened 2018-08-10
Mods for 32 bit Linux

jeanconn:
Update ska3-core to include modifications needed for 32bit linux. 

Also add a small patch for astropy when compiled on ancient 32 bit build system (the conda gcc didn't work either).

### DIFF
--- a/ska_conda/pkg_defs/astropy/meta.yaml
+++ b/ska_conda/pkg_defs/astropy/meta.yaml
@@ -8,7 +8,8 @@ source:
   fn: astropy-{{version}}.tar.gz
   url: https://pypi.io/packages/source/a/astropy/astropy-{{version}}.tar.gz
   sha256: 6af07abe5e263820a3dec93832a6ad74005013071490e125afbc6514411721da
-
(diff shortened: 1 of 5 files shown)


jeanconn:
Looks like this basically works but I don't have a working 32 bit sherpa 4.9.1 (and that build didn't come together trivially).

jeanconn:
I don't know what you mean by "ska3-flight as one-and-only "prime".

Basically, it saves us a bunch of maintenance by having ska3-flight be a noarch package and you can't have different lists by operating system in a noarch package.  So do we keep the 32 bit list separate (new list) or remove sherpa from flight (share list without sherpa) to make that happen?  Or go to an arch-specific ska3-flight that will need rebuilding on each os for each package update?

It starts to be easier to just try to get sherpa to work :smile:

jeanconn:
Or, monster hack, could make a "stub" sherpa 4.9.1 package on 32.

jeanconn:
Duh. I should just move sherpa to ska3-core, the one that is an arch package.

PR #66, opened 2018-08-10
Make acis_taco an arch package

### LINKED ISSUE #57: acis_taco recipe doesn't install make_esaview_data.py

jeanconn:
Remove "noarch: python" from acis_taco recipe so it becomes a regular arch package.

It looks like this is the easiest way to let the current acis_taco install make_esaview_data.py (which it is set to do from setup.py). It also puts the task_sched* files in prefix/share/acis_taco.

Closes #57

### DIFF
--- a/ska_conda/pkg_defs/acis_taco/meta.yaml
+++ b/ska_conda/pkg_defs/acis_taco/meta.yaml
@@ -3,7 +3,6 @@ package:
   version:  {{ GIT_DESCRIBE_TAG }}
 
 build:
-  noarch: python
   script_env:
     - SKA_TOP_SRC_DIR
 


jeanconn:
For the record, the attempt at the noarch package warned about this

```
number of files: 16
Fixing permissions
WARNING:conda_build.noarch_python:Don't know how to handle file: share/make_esaview_data.py.  Omitting it from package.
WARNING:conda_build.noarch_python:Don't know how to handle file: share/acis_taco/make_esaview_data.py.  Omitting it from package.
WARNING:conda_build.noarch_python:Don't know how to handle file: share/acis_taco/task_schedule.cfg.  Omitting it from package.
WARNING:conda_build.noarch_python:Don't know how to handle file: share/acis_taco/task_schedule_occ.cfg.  Omitting it from package.
Fixing permissions
```
and the arch build (on linux-64) does not complain and seems to put at least the files we want in share

```
number of files: 21
Fixing permissions
Fixing permissions
Detected hard-coded path in text file bin/esaview.py
```

PR #67, opened 2018-08-11
Cron runners

jeanconn:
Add recipes for task_schedule and watch_cron_logs

### DIFF
--- a/ska_conda/pkg_defs/task_schedule/build.sh
+++ b/ska_conda/pkg_defs/task_schedule/build.sh
@@ -1,0 +1,4 @@
+export SKA_ARCH_OS=${PREFIX}
+make install3
+
+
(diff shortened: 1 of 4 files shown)


jeanconn:
The tools themselves look to need mods to work in the noarch conda env with modern Perl

jeanconn:
@taldcroft  Do task_schedule and watch_cron_logs belong in ska3-flight as well?  I suppose they won't cause any trouble on a "standalone" style installation.

jeanconn:
But then I need to build them for the other platforms etcs.

jeanconn:
And then we'd basically have perl in ska3-flight which is totally not what we want.  So I'll just leave these as recipes and built packages in skare3 for now, but not as part of any metapackage.

PR #68, opened 2018-08-11
Reorganize ska_conda

jeanconn:
Set ska_builder.py to be a self-contained script .

The ska_conda package is no more.  This reorganizes what was previously "package data" to just make the recipes part of the git repo.  Logic has also been changed to simplify when the git-fetching components are called, and the build list is now an option.


ska_builder.py
```diff
@@ -41,102 +40,98 @@
...
+
+def get_repo(name, tag):
+    # Scan the meta.yaml for GIT_DESCRIBE_TAG and if it isn't there, return None for the,
+    # repo.  That will proceed with a traditional conda build without git fetch etc
+    metayml = os.path.join(pkg_defs_path, name, "meta.yaml")
+    meta = open(metayml).read()
+    has_git = re.search("GIT_DESCRIBE_TAG", meta)
+    if not has_git:
+        return None
+    repo_path = os.path.join(SRC_DR, name)
```
jeanconn: Oops

PR #72, opened 2018-08-13
Add a perlska wrapper

jeanconn:
Add a perlska wrapper

I don't know if we need it, but I'm coming around to the idea that it would make the most sense for the shebang in task_schedule and watch_cron_logs to call a perl wrapper that sets ska.

### DIFF
--- a/pkg_defs/ska3-template/bin/perlska
+++ b/pkg_defs/ska3-template/bin/perlska
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+SKA_bin=`dirname $0`
+eval `${SKA_bin}/flt_envs -shell sh`
+exec ${SKA_bin}/perl "$@"
+
+exit 1
--- a/pkg_defs/ska3-template/build.sh
+++ b/pkg_defs/ska3-template/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-for file in flt_envs ska_envs.sh ska_envs.csh ska_version skare
+for file in flt_envs ska_envs.sh ska_envs.csh ska_version skare perlska
    do
     cp -a ${RECIPE_DIR}/bin/${file} ${PREFIX}/bin
     chmod +x ${PREFIX}/bin/${file}
--- a/pkg_defs/ska3-template/meta.yaml
+++ b/pkg_defs/ska3-template/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: ska3-template
-  version:  2018.08.12
+  version:  2018.08.13
 
 build:
   noarch: generic


pkg_defs/ska3-template/bin/perlska
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+SKA_bin=`dirname $0`
+eval `${SKA_bin}/flt_envs -shell sh`
+exec ${SKA_bin}/perl "$@"
```
jeanconn: This assumes the perl executable is in the same bin, which seems fair.

taldcroft:
> I don't know if we need it, but I'm coming around to the idea that it would make the most sense for the shebang in task_schedule and watch_cron_logs to call a perl wrapper that sets ska.

See comment for task_schedule, but I don't think `perlska` is needed.  I think that the `skare` universal launcher is all that is needed, with the scripts themselves using `#!/usr/bin/env perl/python`.

jeanconn:
OK.  I suppose 50% of the reason I was looking at this is because I can't edit your crontab for twiki-wg and thought mayb perlska might be a more universal way to go.  I'm fine with skare and `#!/usr/bin/env perl"`. 

PS `#!/usr/bin/env python` doesn't actually seem to come up, because conda is more over-eager to rewrite python shebangs whether we want it to or not (flt_envs).

jeanconn:
Closing as unneeded

taldcroft:
>  I suppose 50% of the reason I was looking at this is because I can't edit your crontab for twiki-wg 

My crontab for twiki-wg is no longer active, and you have the action to migrate this over to being run as aca.

taldcroft:
Re: conda editing of python shebangs, I see the following for acis_taco.  Is this not expected for you?  Conda definitely should not be editing these `share` scripts since they are installed as "data" files.

```
ska3-kadi$ head /proj/sot/ska3/flight/share/acis_taco/make_esaview_data.py 
#!/usr/bin/env python
# Licensed under a 3-clause BSD style license - see LICENSE.rst
"""
Calculate Earth illumination on the ACIS radiator over a specified interval
of time.
"""
```

https://github.com/sot/acis_taco/blob/master/setup.py#L16

jeanconn:
Agreed that it didn't edit the shebang of make_esaview_data.py . 

Not sure about when it decides to do the shebang search and replace (because it seemed to just be looking at all the installed files, but maybe it does respect some of the stuff pip has done).   My plan was to watch for this, but also we shouldn't care (if a shebang is rewritten to be explicit, that isn't going to be a problem in a conda env).

PR #75, opened 2018-08-14
Add an aca_egse recipe

jeanconn:
Add an aca_egse recipe


jeanconn:
I think it is non-controversial to add the capability to build the package with this PR.

PR #78, opened 2018-08-19
Add jplephem recipe

jeanconn:
Add jplephem recipe

This doesn't have a built conda package in defaults, so adding the recipe from conda-forge if we choose to add it to skare3.


taldcroft:
Assuming this works and has been tested at some level then :+1:.

jeanconn:
In this case, in my first build, I removed the "--no-test" option to conda build, let the recipe run its own tests (defined in the meta.yaml) and those tests didn't complain.  

Thus far only built/tested on linux-64 but I will check to see if it works on osx and linux-32.  If it just works, I'll configure as such in ska3-core in a separate PR.

PR #87, opened 2018-08-28
Add blas to ska3-core

jeanconn:
Add blas to ska3-core

### DIFF
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: ska3-core
-  version: 2018.08.13
+  version: 2018.08.27
 
 requirements:
   # Packages required to run the package. These are the dependencies that
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -34,6 +34,7 @@ requirements:
    - babel ==2.5.0
    - beautifulsoup4 ==4.6.0
    - bkcharts ==0.2
+   - blas ==1.0
    - bleach ==1.5.0
    - bokeh ==0.12.7
    - certifi ==2016.2.28 # [not linux32]
--- a/pkg_defs/ska3-flight/meta.yaml
+++ b/pkg_defs/ska3-flight/meta.yaml
@@ -9,7 +9,7 @@ requirements:
   # Packages required to run the package. These are the dependencies that
   # will be installed automatically whenever the package is installed.
   run:
-    - ska3-core ==2018.08.13
+    - ska3-core ==2018.08.27
     - ska3-template ==2018.08.12
     - agasc ==3.5.2
     - annie ==0.5
--- a/pkg_defs/ska3-flight/meta.yaml
+++ b/pkg_defs/ska3-flight/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: ska3-flight
-  version: 2018.08.27
+  version: 2018.08.28
 
 build:
   noarch: generic


jeanconn:
I'm not sure how this "went missing".  Looks like numpy 1.12.1 wants it at our current build version so I don't completely understand why it isn't already installed or why the same version of conda now solves the dependency tree and wants to install it.

This should get some more testing (ska_testr etc).

taldcroft:
Hmm, that is indeed mysterious.  I poked around at this for a bit in my inexpert way and don't see a resolution.  It looks like this exact numpy build just accrued a new dependency on blas.  I'm comparing the json info file I have on disk in conda-meta with `conda info numpy`, where I have set up my .condarc to the channels for building.
```
(ska3) neptune$ more numpy-1.12.1-py36_0.json 
{
  "arch": "x86_64",
  "build": "py36_0",
  "build_number": 0,
  "channel": "https://repo.anaconda.com/pkgs/free/osx-64",
  "constrains": [],
  "date": "2017-03-18",
  "depends": [
    "mkl 2017.0.*",
    "python 3.6*"
  ],
 ...
  "md5": "dbad85b87520e45987d6e14011ef4489",
```

```
numpy 1.12.1 py36_0
-------------------
file name   : numpy-1.12.1-py36_0.tar.bz2
name        : numpy
version     : 1.12.1
build string: py36_0
build number: 0
channel     : https://repo.continuum.io/pkgs/free/osx-64
size        : 3.5 MB
arch        : x86_64
constrains  : ()
date        : 2017-03-18
license     : BSD
md5         : dbad85b87520e45987d6e14011ef4489
platform    : darwin
subdir      : osx-64
url         : https://repo.continuum.io/pkgs/free/osx-64/numpy-1.12.1-py36_0.tar.bz2
dependencies:
    blas * mkl
    mkl 2017.0.*
    python 3.6*
```

It's picking up the `blas` package from `defaults`, which doesn't seem right to me:
```
(ska3) neptune$ conda install --dry-run  numpy=1.12.1
Solving environment: done

## Package Plan ##

  environment location: /Users/aldcroft/miniconda3/envs/ska3

  added / updated specs: 
    - numpy=1.12.1


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    blas-1.0                   |              mkl           5 KB  defaults
    openssl-1.0.2l             |                0         3.0 MB  defaults
    ------------------------------------------------------------
                                           Total:         3.0 MB

The following NEW packages will be INSTALLED:

    blas:    1.0-mkl           defaults                           

The following packages will be DOWNGRADED:

    openssl: 1.0.2l-h492a243_5 https://repo.anaconda.com/pkgs/main --> 1.0.2l-0 defaults


DryRunExit: Dry run exiting
```
```
$ tail ~/.condarc

#channels:
#  - defaults
#  - https://icxc.cfa.harvard.edu/aspect/ska3-conda

binstar_upload: false
channels:
  - https://repo.continuum.io/pkgs/free
  - https://repo.continuum.io/pkgs/r
  - https://repo.continuum.io/pkgs/msys2

jeanconn:
I tried a similar experiment, but I ran into the issue that if I downloaded the current numpy tar.gz (ignoring the fact that it shouldn't have had an update because the date is old) and untarred it, the index json inside has the same "depends" as the file we have.  So I don't immediately see how the package itself can be "asking" for the blas.  I don't understand why the dependency solver and conda info seem to have the blas dependence.


> It's picking up the blas package from defaults, which doesn't seem right to me:

I'm not sure about the "rightness" of that.  I think the 'defaults' channel is special, and with the edit you just did (replacing 'default_channels' with channels' to try to have channels that don't include defaults) you've still got the built-in defaults in the search path.  See "conda config --show" ?  I think if you explicitly want to reduce the set of channels to just the three you have (which we've been doing via default_channels) you could also set default_channels to the empty array (default_channels: [] ).  Or at the command line use '--override-channels'.

I'm guessing that defaults is going to be included by default but that we list it in 'channels' to try to define priority explicitly.

jeanconn:
I see you approved this PR... how much do you want to dig?  I can open issues on anaconda and the conda mailing list, but another problem with running ancient versions of the tools to support CentOS5 is that I suspect I'm just going to get "update conda" as feedback.

PR #88, opened 2018-08-31
ACIS updates approved 31-Aug-2018

jeanconn:
This includes changes to cmd_states and kadi to support that WSVIDALLDN sets ccd_count to 0.
There's also a new version of acis_thermal_check that uses the sqlite database by default instead of Sybase.


jeanconn:
I've tacked on the kadi test fix here, though that was not explicitly brought to load review with the rest on Aug 31

PR #89, opened 2018-10-04
Small updates pre proseco

jeanconn:
Small updates pre proseco

### DIFF
--- a/pkg_defs/ska3-flight/meta.yaml
+++ b/pkg_defs/ska3-flight/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - chandra_aca ==4.23
     - chandra.cmd_states ==3.14.2
     - chandra.maneuver ==3.7.1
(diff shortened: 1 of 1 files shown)


jeanconn:
I think we just want the chandra_aca update and then this is done.

jeanconn:
Unless we want to put newer annie in or proseco 4.0 while we're at it.

taldcroft:
Don't need annie or proseco.

jeanconn:
We already have an annie 0.5 in ska3-flight.  I was just wondering when to bother to update to 0.6.

taldcroft:
I don't think we should release a new version until @malgosias has reviewed and merged https://github.com/sot/annie/pull/47.

jeanconn:
Any opinion on just tossing jplephem into ska3-flight?  It is a noarch package at this point.

taldcroft:
Jplephem sounds good.

PR #96, opened 2018-12-14
Update ska_sync to 4.2

taldcroft:
To be pulled in with the next release.

### DIFF
--- a/pkg_defs/ska3-flight/meta.yaml
+++ b/pkg_defs/ska3-flight/meta.yaml
@@ -44,7 +44,7 @@ requirements:
     - ska.matplotlib ==3.11.2
     - ska.parsecm ==3.3.1
     - ska_path ==3.1
-    - ska_sync ==0.1.1
+    - ska_sync ==4.2
     - ska.quatutil ==3.3.1
     - ska.shell ==3.3.3
     - ska.sun ==3.5


jeanconn:
I need a ska3-flight with proseco 4.3.1 whether it goes into HEAD ska3 or not (I think we always want the matlab tools installed from a ska3-flight or a new metapackage).  So I can bring to load review or not.

Do you want me to just toss in this ska_sync update?  I have not prepared to brief.

taldcroft:
Just toss this in, no need to brief.  You have FD signoff for this.

jeanconn:
OK.  In general, I think we want a "ska3-flight" that corresponds to the proseco version, but since we don't need to install it on HEAD, I don't know if the proseco side needs brief either.

taldcroft:
Being lazy and minimalist I'm not sure I would have cut a new ska3-flight for this.  We need the new package for ska3-matlab32, but I don't know if anybody will call ska3-flight/proseco 4.3.1 just yet.

But it doesn't hurt, only about your effort.

jeanconn:
That's the issue right now, we don't have a ska3-matlab metapackage and the impression I had from you was that ska3/matlab should just end up on an older version of ska3-flight instead of having its own branched metapackage.

jeanconn:
So if I want to install proseco 4.3.1 on ska3/matlab/test now, I can either uninstall ska3-flight and install proseco 4.3.1, or I can create a new ska3-flight to which to update.

jeanconn:
That's why I don't care about briefing it.  ska3-flight 12.18.2018 could end up being a GRETA only release.

jeanconn:
But let me know if we require a ska3-matlab metapackage and we can construct one (it would look an awful lot like ska3-flight).

taldcroft:
Ah, I had thought there was a ska3-matlab package.  (Sorry, the 32 was a mistake).  Here are the bits from the Ska3 planning doc relevant to ska3-matlab:
- Because of the slower release cycle of Matlab tools and different requirements, this branch will typically lag Flight3.  [Branch is not a github branch here]
- Current process has been to defer updates to the Ska Matlab2 environment until there is a actual driver.  We can stick with this, but the goal with Skare3 (with conda packaging support) is to make the update process much easier.

So indeed I think that ska3-matlab should generally follow ska3-flight (eventually), but ska3-matlab will only get updated in discrete jumps with each MATLAB tools release so that at any point in time it matches the current FDB-approved environment.  That's why a ska3-matlab metapackage is needed, so you can always install either ska3-flight (which will update almost weekly) or ska3-matlab which updates every 2-3 months or so.

jeanconn:
Ah.  OK.  I thought that ska3-matlab was a concept, not a branch or metapackage.  I'll make it a metapackage in the new year.  I think we can manage with this one extra ska3-flight release this week, and I can just brief it to save confusion.

PR #101, opened 2019-01-05
Make ska3-matlab metapackage

jeanconn:
Make a ska3-matlab metapackage.  This starts from ska3-flight 2018.12.18 plus some anticipated bugfixes.

### DIFF
--- a/pkg_defs/ska3-matlab/meta.yaml
+++ b/pkg_defs/ska3-matlab/meta.yaml
@@ -37,7 +37,7 @@ requirements:
     - ska.arc5gl ==3.1.1
     - ska.astro ==3.2.1
     - ska.dbi ==3.8.2
-    - ska.engarchive ==4.44
+    - ska.engarchive ==4.44.1
     - ska.file ==3.4.1
     - ska.ftp ==3.5
     - ska.numpy ==3.8.1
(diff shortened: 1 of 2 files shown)


jeanconn:
Sure seems like ska3-flight 2018.12.18 is fine if you want to match what is installed right now on GRETA (IIRC, haven't logged in over there today to check).

Anyway, with regard to keeping ska3/matlab32 up-to-date, I assume that ska3/matlab32 should get replaced to match the directory structure on GRETA, but there were a few other things going on at the time.

jeanconn:
IIRC I also gave up on trying to test the environment on HEAD and moved to just testing on GRETA; but if you think there's still value in using quango as a proxy for their machines that's fine.

jeanconn:
PS With regard to ska3/matlab being ancient... I think it is only outdated by 5 Jean-in-the-office days :smile:.

taldcroft:
Good point about using GRETA as the test platform 32-bit since it is really the only target.  And with that we could perhaps just not have /proj/sot/ska3/matlab32 on HEAD.  From my perspective better to not have it all than potentially be incorrect.

I put some new details in the [package developer process](https://github.com/sot/skare3/wiki/Ska3-runtime-environment-for-maintainers#package-developer-process).

PR #114, opened 2019-02-20
Fix preprocessing selector logic bugs in ska3-core

### LINKED ISSUE #111: Could use git on GRETA
Should git be part of ska3-matlab metapackage or should it just be conda installed in no metapackage... or something else?

jeanconn:
Fix preprocessing selector logic bugs

All of the "and" statements were wrong as when building a platform cannot both be linux64 and osx.  I think this was just not noticed as most of these explicit package includes get pulled in by other packages.  The incorrect boolean logic resulted in the lines just being silently dropped when the linux64 and osx metapackages were built.

Closes #111

### DIFF
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: ska3-core
-  version: 2018.08.27
+  version: 2019.02.20
 
 requirements:
   # Packages required to run the package. These are the dependencies that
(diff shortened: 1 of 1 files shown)


jeanconn:
@mbaski I just noticed that the packages in this diff are not explicitly installed at the right versions in the current GRETA Linux 64 bit test environment (which isn't really used for anything anyway).  It is my plan to fix the metapackage and install/update these 7 packages on linux64 as part of this next MATLAB release. 

This should just be a no-op on the Windows side.

PR #129, opened 2019-04-11
Add python-pptx recipe and add it to ska3-core on linux64

jeanconn:
Add python-pptx recipe and add it to ska3-core on linux64

### DIFF
--- a/pkg_defs/python-pptx/build.sh
+++ b/pkg_defs/python-pptx/build.sh
@@ -1,0 +1,2 @@
+export SKA=/dev/null
+pip install --no-deps --verbose --no-binary :all: --no-index .
--- a/pkg_defs/python-pptx/meta.yaml
+++ b/pkg_defs/python-pptx/meta.yaml
@@ -1,0 +1,53 @@
+{% set version = "0.6.17" %}
+
+package:
+  name: python-pptx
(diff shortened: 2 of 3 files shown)


taldcroft:
For the record can you remind me where to find the build recipe?  I can find everything but that on conda-forge.

jeanconn:
There is no one place.  For this package that has a conda-forge recipe, I just stole and modified from https://github.com/conda-forge/python-pptx-feedstock/blob/master/recipe/meta.yaml

taldcroft:
And how did you discover that, and what is your general discovery process.  All this should go into the maintainer docs so we aren't single strung on all the conda stuff.

jeanconn:
My approach was to google.  But that seems poor for the docs :smile: .

jeanconn:
I think we decided against including this package in ska3-flight, but I'm not sure if there is value in just merging the recipe (per the approval).

jeanconn:
I wouldn't want to merge the update to ska3-core.

taldcroft:
Agreed to merge the recipe but take it out of ska3-core.  I am not sure why I approved this, maybe didn't appreciate the inclusion in ska3-core when I clicked approve.

PR #132, opened 2019-05-23
Add astropy-healpix recipe

jeanconn:
As noted in recipe comment, this recipe largely borrowed from

https://github.com/conda-forge/astropy-healpix-feedstock

but the build and run requirements have been set manually to ska3 requirements.

### DIFF
--- a/pkg_defs/astropy-healpix/meta.yaml
+++ b/pkg_defs/astropy-healpix/meta.yaml
@@ -1,0 +1,54 @@
+# Recipe largely borrowed from https://github.com/conda-forge/astropy-healpix-feedstock
+# but the build and run requirements have been set manually to ska3 requirements
+
+{% set name = "astropy-healpix" %}
+{% set version = "0.4" %}
+{% set sha256 = "8c9709ac923759c92eca6d2e623e734d0f417eed40ba835b77d99dec09e51aa2" %}
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 1001
+  script: python setup.py install --single-version-externally-managed --record record.txt --offline --no-git
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - numpy ==1.12.1
+    - llvmlite ==0.18.0
+    - gcc ==4.8.5 # [linux64]
+    - libgcc ==4.8.5 # [linux64]
+    - astropy
+  run:
+    - python
+    - llvmlite ==0.18.0
+    - numpy ==1.12.1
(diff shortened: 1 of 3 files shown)


pkg_defs/astropy-healpix/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - gcc ==4.8.5 # [linux64]
+    - libgcc ==4.8.5 # [linux64]
+    - astropy
+  run:
+    - python
+    - llvmlite ==0.18.0
+    - numpy ==1.12.1
+    - astropy >=2.0
+    - libgcc ==4.8.5
+    - astropy >=2.0
```
taldcroft: Repeat.

pkg_defs/astropy-healpix/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+build:
+  number: 1001
+  script: python setup.py install --single-version-externally-managed --record record.txt --offline --no-git
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - numpy ==1.12.1
+    - llvmlite ==0.18.0
```
taldcroft: Does not matter that much, but I don't understand this dependency.

https://astropy-healpix.readthedocs.io/en/latest/installation.html#required-dependencies

jeanconn: This is more me throwing in our ancient llvmlite on anything that might use gcc or libgcc, as the pin doesn't seem respected in a build context.

pkg_defs/astropy-healpix/meta.yaml
jeanconn: Oops, I cut the wrong one to have github pick this up as outdated.

pkg_defs/astropy-healpix/meta.yaml
taldcroft: Can you add a comment to that effect?

jeanconn: Sure.  Though the recipe saved in the now-built version of astropy-healpix won't have it.

jeanconn:
This is build in the flight channel now.

pkg_defs/astropy-healpix/meta.yaml
taldcroft: The new comment does not capture the confusing part that you are apparently pinning gcc by including and pinning an entirely unrelated and unused package llvmlite.

jeanconn: No, I'm trying to keep binary compatibility by pinning llvmlite directly and we also need to build with the old compiler.  What would you like the comment to say?  Mostly I just keep adding things to the recipe until the errors go away and I haven't tried dig in as much into the "why" with the hope that all of these explicit versions related to gcc/llvmlite will go away.

taldcroft: Well I'm just lost because llvmlite is not a dependence of this package AFAIK so I have no clue why it appears in the recipe.  But I guess it doesn't matter you will remember to remove it later.

PR #136, opened 2019-05-24
Update recipe for xija with numba and pinned llvmlite

jeanconn:
Update recipe for xija with numba and pinned llvmlite


jeanconn:
This is built in the flight channel now.

PR #153, opened 2019-07-01
Update ska3-core and ska3-flight versions for 2019.07.01 release

jeanconn:
Update ska3-core and ska3-flight versions for 2019.07.01 release


jeanconn:
I think it makes some sense to just have the version updates for the metapackages ride along in a separate PR for now.

PR #156, opened 2019-07-02
Add conda itself (at version 4.3.21) to ska3-core

jeanconn:
Add conda itself (at version 4.3.21) to ska3-core


jeanconn:
I think my original thinking was that this only needed to be defined explicitly in the ska_builder package.  We generally don't install that package in ska3-flight or ska3-matlab, however, so the conda version keeps wanting to creep.  I think this is a reasonable way to just set its version.

PR #165, opened 2019-07-17
Update ska3-flight and ska3-core versions for 2019.07.15 weekly release

jeanconn:
Update ska3-flight and ska3-core versions for 2019.07.15 weekly release


jeanconn:
These aren't needed; no update installed during the 2019.07.15 week.

PR #166, opened 2019-07-17
Update mica to 4.18

### LINKED ISSUE #150: Add mica with starcheck cron task (vers 4.18) to ska3-flight

jeanconn:
Update mica to 4.18

Closes #150

### DIFF
--- a/pkg_defs/mica/meta.yaml
+++ b/pkg_defs/mica/meta.yaml
@@ -27,7 +27,7 @@ requirements:
     - python
     - six
     - matplotlib
-    - numpy
+    - numpy ==1.12.1
     - pytest
     - jinja2
     - astropy
--- a/pkg_defs/mica/meta.yaml
+++ b/pkg_defs/mica/meta.yaml
@@ -3,7 +3,7 @@ package:
   version:  {{ GIT_DESCRIBE_TAG }}
 
 build:
-  noarch: python
+  noarch: generic
   script_env:
     - SKA_TOP_SRC_DIR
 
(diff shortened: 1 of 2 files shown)


jeanconn:
I added numpy 1.12 here because now that I'm trying to remove the "defaults" from the .condarc, this wasn't building.  For some reason it was trying to build using numpy 1.11 which is not in our repo, even though a perfectly useful numpy 1.12.1 is in our repo.  I think a bunch of these problems will go away with newer conda; and I'll need to come back and fix recipes at that point.

jeanconn:
It looks like I'm not getting the task_schedule this way though

WARNING:conda_build.noarch_python:Don't know how to handle file: share/mica/task_schedule.cfg.  Omitting it from package.

The acis_taco package that appeared to be the model for this is an arch package (not a no-arch python conda package) so that's handled differently.  If I can't figure out a work around we can just convert mica to an arch package (though that will be annoying for basically pure Python packages).

jeanconn:
It looks like converting from `noarch: python` to `noarch: generic` seems to work?

The mica package made noarch generic looks to install like a Python package but also keeps the task schedule and scripts.  The internals of the conda package look a little different, but I don't think we care?  I've installed in my devska and the mica tests pass; do we need to do other testing for that change?  I can't find useful documentation describing what "magic" we're not getting if we don't use `noarch: python` on a python package.

jeanconn:
Otherwise I could just remove the `noarch` altogether and have this install like acis_taco.

PR #171, opened 2019-08-06
Move starcheck pkg into ska3-flight

jeanconn:
Remove some starcheck dependencies from recipe

This should allow the starcheck conda package to build successfully and without the Perl dependencies.  Installing that package should make starcheck.parser available for proseco and annie without requiring a working starcheck/Perl environment.

Closes https://github.com/sot/mica/issues/204

### DIFF
--- a/pkg_defs/ska3-flight/meta.yaml
+++ b/pkg_defs/ska3-flight/meta.yaml
@@ -51,6 +51,7 @@ requirements:
     - ska.sun ==3.5
     - ska.tdb ==3.5.1
     - sparkles ==4.3
+    - starcheck ==13.4
     - tables3_api ==0.1
     - testr ==3.2
     - xija ==4.15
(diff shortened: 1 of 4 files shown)


jeanconn:
@taldcroft And I think these are the rest of the skare3 changes needed so we can install starcheck as a Python package to get the parser, but it won't be fully functional until the perl pieces are installed in the environment.  With these changes the entry for the starcheck package should also be able to move from ska3-perl to ska3-flight.

jeanconn:
I suppose in skare3 I'm somewhat loathe to promote this, as we generally build out of master and this would build a different ska3-flight with the same version as installed.  I'm not sure about the best safeguards for that, except being clear to either:

1) only build the individual package you mean to install
2) only promote / install the individual package you've just built to the conda channel

The rsync's are too sloppy.

taldcroft:
> I suppose in skare3 I'm somewhat loathe to promote this, as we generally build out of master and this would build a different ska3-flight with the same version as installed. 

I'm lost.  My understanding is that we *always* build packages from a tagged release, end of story.  

Beyond that, you've made a large leap from this particular PR to a bigger picture question and you need to fill in a lot of details for me, maybe with a specific example and more words.

PR #173, opened 2019-08-08
Remove nbpresent from ska3-core

### LINKED ISSUE #90: Remove nbpresent from ska3-core
The nbpresent jupyter extension is nothing but an annoyance to me (it shows a useless toolbar at the bottom and seems to take away the very useful toolbar at the top).  Since we never do notebook presentations, I suggest removing this from our core package list.

taldcroft:
This package is mostly just annoying as it makes it easy to lose the jupyter top menu bar in favor of the bottom nbpresent bar that we never use.

Closes #90

### DIFF
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -108,7 +108,6 @@ requirements:
    - nb_conda_kernels ==2.1.0
    - nbconvert ==5.2.1
    - nbformat ==4.4.0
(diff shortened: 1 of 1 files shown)


jeanconn:
Oops.  It looks like this wasn't tested, so I'm reverting it for now, my mistake.  Removing nbpresent from ska3-core doesn't actually remove the package (there's an interaction with _nb_ext_conf?).  If I remove that too, what do we need to test still works in jupyter notebooks?

PR #183, opened 2019-08-13
Update flt_envs for Sybase 16.0

jeanconn:
Update flt_envs for Sybase 16.0


jeanconn:
On the Python side, this seems to work fine with just these two changes (LD_LIBRARY_PATH just seems to need the previous single lib directory).

Need to test Perl side.

PR #191, opened 2019-08-20
Update recipes to allow some ska packages to use more modern numpy

jeanconn:
Update recipes to allow some ska packages to use more modern numpy

It would be great if we can make these recipes successfully build on CentOS5 but make packages that have requirements such that they can work on CentOS7 and OSX in a ska3-dev environment with more modern numpy (for testing and dev work).

There are still other packages that have numpy 1.12 hardcoded requirements (astropy, jplephem) but my hope would be that we could use those packages from a more modern repo from continuum in the more modern environment, instead of needing to build them in skare3.

### DIFF
--- a/pkg_defs/mica/meta.yaml
+++ b/pkg_defs/mica/meta.yaml
@@ -3,7 +3,6 @@ package:
   version:  {{ GIT_DESCRIBE_TAG }}
 
 build:
-  number: 1
   noarch: generic
   script_env:
     - SKA_TOP_SRC_DIR
--- a/pkg_defs/mica/meta.yaml
+++ b/pkg_defs/mica/meta.yaml
@@ -3,6 +3,7 @@ package:
   version:  {{ GIT_DESCRIBE_TAG }}
 
 build:
+  number: 1
   noarch: generic
   script_env:
     - SKA_TOP_SRC_DIR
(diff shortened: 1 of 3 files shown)


jzuhone:
Because we pin to NumPy 1.12 in ska3-flight, this change doesn't affect that, right? Only ska-dev. (which is what we want for now)

jeanconn:
Right (though technically I think we set the version explicitly in ska3-core which itself is explicitly versioned in ska3-flight).  I have just had issues with getting these packages to *build* correctly against numpy 1.12 (what we want now) without having the explicit requirement, so these requirements crept into the recipes.  

As I was saying in the PR top, it would be nice to have recipes that still work on the old build system, but if we need to make a new branch to store the updated recipes, that could work as well and this could be a start.

jeanconn:
It looks like these are building fine now because the build system is only using our core-pkg-repo as a source of packages (so it is just finding the old numpy 1.12 and it isn't trying to build the 32 bit package with numpy 1.16 built with modern compilers and then dying).  @taldcroft  do you think this could actually just go today and be used to build the 4.15 xija package?

jeanconn:
Right, the >= 1.12.1 was because strangely it was complaining about a missing numpy 1.11 when trying to build previously.  I'll double-check it is still a problem;  I don't know how the solver in our old version was deciding that was a good idea.

The "number: 1" just labels these as "build number 1" in the built packages so they technically could be distinguished from the previous packages at the same module version.  I can cut that if we don't care.

jeanconn:
I'll cut the build number updates (I suppose really just useful during testing) and merge this.

PR #206, opened 2019-09-20
Ska/Python updates in support of Matlab 2019_205

jeanconn:
Ska/Python updates in support of Matlab 2019_205

### DIFF
--- a/pkg_defs/ska3-matlab/meta.yaml
+++ b/pkg_defs/ska3-matlab/meta.yaml
@@ -36,7 +36,7 @@ requirements:
     - ska.arc5gl ==3.1.1
     - ska.astro ==3.2.1
     - ska.dbi ==3.8.2
-    - ska.engarchive ==4.44.1
+    - ska.engarchive ==4.45
     - ska.file ==3.4.1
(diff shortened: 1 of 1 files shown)


pkg_defs/ska3-matlab/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: ska3-matlab
-  version: 2019.06.13
+  version: 2019.10.02
```
taldcroft: Would it make sense to make this version number be the same as the MATLAB release, `2019_205`?  I was also wondering about release tags like `MATLAB_2019_205`.

jeanconn: I like the idea, but how does that work if we don't have the final package list in our "first go"?  With an independent version, if I add a package or bugfix during our process I can just update this version and have a unique label associated with the updated package list.  If we use a dependent version like `205`  I could relabel just before final testing, but that would be extra package making/installing/testing.

I do consider these versions like public tags in that if if I use one I *really really* try not to reuse the same version for a new package with different contents.

PR #217, opened 2019-11-01
fixed acisfp_check build.sh.

javierggt:
Defined SKA=/proj/sot/ska and this worked in the build environment. Otherwise, the error was "no local Ska data found and remote access is not selected"


jeanconn:
This patch to just build this on presto works for me; though we'll need to fix again if we want to successfully build this  noarch package on zero / oxs build machine.

PR #227, opened 2019-11-22
WIP windows conda build for ska3-win

taldcroft:
This is a proof-of-concept for making Windows builds which seems to be working.  I had quite a journey getting this working so ended up doing micro-commits to allow backtracking to a configuration that works.

It seems that using `python setup.py install` in this context is fine since conda is doing the file management and so `pip` is a bit of overkill.  But perhaps more importantly, modern `pip` has dropped support for eggs and this is needed for namespace packages.  So going to `python setup.py install` should allow upgrading `pip` to the latest.

It also seems like we can just use the `build : script` directive in `meta.yaml`, but maybe there is a reason not to that @jeanconn has discovered.

Next steps would be doing the rest of the arch-specific packages and then dealing with the core packages.  I'm guessing that our totally-pinned ska3-core package won't resolve on Windows, but not sure.

There is something a bit unsettling in the build log.  I did some googling and got to some relevant pages, but couldn't quite make sense of them.
```
Packaging chandra.time
INFO:conda_build.build:Packaging chandra.time
INFO conda_build.build:build(1570): Packaging chandra.time
Packaging chandra.time-3.20.3-py36_0
INFO:conda_build.build:Packaging chandra.time-3.20.3-py36_0
INFO conda_build.build:bundle_conda(891): Packaging chandra.time-3.20.3-py36_0
compiling .pyc files...
number of files: 22
   INFO (chandra.time,Lib\site-packages\Chandra.Time-3.20.3-py3.6-win-amd64.egg\Chandra\Time\_axTime3.cp36-win_amd64.pyd): Needed DSO python36.dll found in defaults::python-3.6.9-h5500b2f_0
WARNING (chandra.time,Lib\site-packages\Chandra.Time-3.20.3-py3.6-win-amd64.egg\Chandra\Time\_axTime3.cp36-win_amd64.pyd): Needed DSO api-ms-win-crt-string-l1-1-0.dll found in ['vs2015_runtime']
WARNING (chandra.time,Lib\site-packages\Chandra.Time-3.20.3-py3.6-win-amd64.egg\Chandra\Time\_axTime3.cp36-win_amd64.pyd): .. but ['vs2015_runtime'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
[...]

PR #230, opened 2019-11-28
WIP Shiny packages updates (CentOS6 conda 4.7 so far)

jeanconn:
Recipe changes to start to build modern packages with conda 4.7.

This was *just* building and testing on CentOS6 .

### DIFF
--- a/pkg_defs/Chandra.Maneuver/build.sh
+++ b/pkg_defs/Chandra.Maneuver/build.sh
@@ -1,1 +1,1 @@
-pip install --no-deps --verbose --no-binary :all: --no-index --egg .
+python setup.py install --old-and-unmanageable
--- a/pkg_defs/Chandra.Time/meta.yaml
+++ b/pkg_defs/Chandra.Time/meta.yaml
@@ -6,6 +6,7 @@ build:
   script_env:
     - SKA_TOP_SRC_DIR
   preserve_egg_dir: yes # (default no)
+  script: python setup.py install
 
 source:
   path: {{ SKA_TOP_SRC_DIR }}/Chandra.Time
--- a/pkg_defs/Chandra.Time/meta.yaml
+++ b/pkg_defs/Chandra.Time/meta.yaml
@@ -2,16 +2,13 @@ package:
   name: chandra.time
   version:  {{ GIT_DESCRIBE_TAG }}
 
-#build:
-# script_env:
-#   - SKA_TOP_SRC_DIR
-# preserve_egg_dir: yes # (default no)
+build:
+  script_env:
+    - SKA_TOP_SRC_DIR
+  preserve_egg_dir: yes # (default no)
 
 source:
   path: {{ SKA_TOP_SRC_DIR }}/Chandra.Time
-  #git_rev: "3.20.3"
-  #git_url: https://github.com/sot/chandra.time
-
 
 # the build and runtime requirements. Dependencies of these requirements
 # are included automatically.
--- a/pkg_defs/Chandra.Time/meta.yaml
+++ b/pkg_defs/Chandra.Time/meta.yaml
@@ -2,13 +2,15 @@ package:
   name: chandra.time
   version:  {{ GIT_DESCRIBE_TAG }}
 
-build:
-  script_env:
-    - SKA_TOP_SRC_DIR
-  preserve_egg_dir: yes # (default no)
+#build:
+# script_env:
+#   - SKA_TOP_SRC_DIR
+# preserve_egg_dir: yes # (default no)
 
 source:
-  path: {{ SKA_TOP_SRC_DIR }}/Chandra.Time
+  #path: {{ SKA_TOP_SRC_DIR }}/Chandra.Time
+  git_rev: "3.20.3"
+  git_url: https://github.com/sot/chandra.time
 
 
 # the build and runtime requirements. Dependencies of these requirements
@@ -18,10 +20,9 @@ requirements:
   # listed explicitly if they are required.
   build:
     - python
-    - pip <10.0
     - cython >=0.20.1
     - six
-    - numpy
+    - numpy =1.12
   # Packages required to run the package. These are the dependencies that
   # will be installed automatically whenever the package is installed.
   run:
--- a/pkg_defs/Chandra.Time/meta.yaml
+++ b/pkg_defs/Chandra.Time/meta.yaml
@@ -8,9 +8,9 @@ package:
 # preserve_egg_dir: yes # (default no)
 
 source:
-  #path: {{ SKA_TOP_SRC_DIR }}/Chandra.Time
-  git_rev: "3.20.3"
-  git_url: https://github.com/sot/chandra.time
+  path: {{ SKA_TOP_SRC_DIR }}/Chandra.Time
+  #git_rev: "3.20.3"
+  #git_url: https://github.com/sot/chandra.time
 
 
 # the build and runtime requirements. Dependencies of these requirements
--- a/pkg_defs/Chandra.Time/meta.yaml
+++ b/pkg_defs/Chandra.Time/meta.yaml
@@ -3,8 +3,6 @@ package:
   version:  {{ GIT_DESCRIBE_TAG }}
 
 build:
-  script_env:
-    - SKA_TOP_SRC_DIR
   preserve_egg_dir: yes # (default no)
   script: python setup.py install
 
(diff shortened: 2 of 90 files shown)


pkg_defs/Chandra.Maneuver/build.sh
```diff
@@ -1 +1 @@
-pip install --no-deps --verbose --no-binary :all: --no-index --egg .
+python setup.py install --old-and-unmanageable
```
jeanconn: I assume we could also just set a standard build in a environment variable.

taldcroft:
I had imagined following the lead of #227 with simply: `script: python setup.py install`  in the `build:` section of `meta.yaml` and get rid of `build.sh`.  This relies on conda to take away the egg dir by default, which is fine, and has the advantage that we will be already set up for Windows builds.  When we migrate all the namespace packages to take out the `Ska/__init__.py` files, then we can get rid of the `preserve_egg_dir` directives, and achieve minimal meta.yaml files.

pkg_defs/Ska.arc5gl/meta.yaml
```diff
@@ -18,7 +18,6 @@ requirements:
   # Packages required to build the package. python and numpy must be
   # listed explicitly if they are required.
   build:
-    - pip <10.0
     - python
     - setuptools
```
taldcroft: While you are at it I propose to move all the explicit `setuptools` dependencies.  Yes, they are a dependence, but this is *always* provided by `python` so this is superfluous.  There are a number of other packages also provided by conda `python` without which everything will break, but we don't explicitly specify those.

jeanconn:
Sure, it was just easier to batch update the  build.sh files for a first go, and as far as I could tell there were a few different common 'python setup.py install' idioms to speed it up for conda use.

pkg_defs/annie/meta.yaml
```diff
@@ -41,7 +40,7 @@ requirements:
     - ska.engarchive
     - mica
     - testr
-    - llvmlite ==0.18.0
+
```
taldcroft: Extra blank line.

pkg_defs/kadi/meta.yaml
```diff
@@ -30,7 +30,7 @@ requirements:
     - configobj
     - requests
     - tables3_api
-    - django <2.0
+    - django
```
taldcroft: Letting `django` float (to 2.2) requires https://github.com/sot/kadi/pull/143 and a re-build of `events.db3`.  I think we can and should go in that direction, but commenting that I believe this won't work out of the box.

ska_builder.py
```diff
@@ -123,7 +122,7 @@ def build_package(name):
         cmd_list.append("--skip-existing")
 
     print('Running:\n' + ' '.join(cmd_list) + '\n')
-    subprocess.run(cmd_list, check=True, shell=True)
+    subprocess.run(cmd_list, check=True)
```
taldcroft:
```
subprocess.run(cmd_list, check=True, shell=(os.name == 'nt'))
```

pkg_defs/annie/build.sh
```diff
@@ -1,2 +1,2 @@
 export SKA=/dev/null
```
taldcroft: Can we instead just ensure that the `SKA` is set in `ska_builder.py` i.e. if currently undefined then set to `/proj/sot/ska`.  Currently `cheta` is strict and won't import without finding an archive, while `mica` will import as long as `SKA` is defined to something.  This is somewhat a philosophical discussion about whether a package should import if it does not have the data to actually function, and perhaps situations like this show the way (contrary to my past thoughts).

The right answer from the packaging perspective is moving to single-source version so `setup.py` can run without importing the package, but in the meantime I think we can clean up these recipes assuming integration with `ska_builder.py` (which is already the case with the `SKA_TOP_SRC_DIR`).

pkg_defs/kadi/meta.yaml
jeanconn: I thought the modern django was actually a requirement for modern dependencies to work, so I was building kadi out of that branch to start.

taldcroft:
Indeed changing to use `python setup.py install` in `meta.yaml` will require writing a little script, but it shouldn't be too bad.  I think it's just looking for the first blank line after `build:` and inserting `'  script: python setup.py install'`.  Good news is that you don't distinguish between namespace packages since that is already done everywhere with the `preserve_egg_dir` (as needed).

Is this `script_env: - SKA_TOP_SRC_DIR` still needed anywhere?  Even now I don't see any build scripts that actually use it.

pkg_defs/annie/meta.yaml
```diff
@@ -23,7 +23,6 @@ requirements:
     - chandra_aca
     - mica
     - testr
-    - llvmlite ==0.18.0
```
taldcroft: :tada:

pkg_defs/kadi/meta.yaml
taldcroft: I'm not sure if any of the rest of our OTS stack depends on django, but who knows.  In any case I'm not arguing against the latest django, just giving a heads-up that it will break kadi hard without a database migration.

taldcroft: Actually it's worse than just the database, IIRC there are changes that won't allow kadi to import.

pkg_defs/annie/build.sh
jeanconn: I was a little concerned, since our default build will likely be on a Linux machine that will see the real /proj/sot/ska and run as the aca user, that giving any more recipes access to /proj/sot/ska increased risk.

taldcroft: ~~Fair enough to not change the ska_builder.py script right now, instead just use `script env SKA=whatever python setup.py install`.~~

Nope, that won't work on Windows.  So back to setting `SKA` to something in ska_builder.py.  Honestly I'm not that concerned about `python setup.py install` ever *writing* to something in SKA if it exists.  What happens there is importing various packages (primarily to get the version), and I always consider it safe to import any Ska package as `aca`.  If the act of importing a package can corrupt our flight database then we have bigger problems.

Did you have some specific failure path in mind that I'm missing?

Nevertheless we can be super careful just the same in ska_builder.py by making a temp dir, point SKA at that, and then make a dir SKA/eng_archive/data.  That will satisfy cheta and mica to get them to import.

pkg_defs/Chandra.Time/meta.yaml
```diff
@@ -17,10 +15,12 @@ requirements:
   # Packages required to build the package. python and numpy must be
   # listed explicitly if they are required.
   build:
+    - {{ compiler("cxx") }} # [ unix ]
+  host:
```
jeanconn: Though now that I've fiddled with this some more, it will need Windows testing.

taldcroft:
> When you say "is there a backstory here" I'm not sure what you mean? I was just trying to use the right options with the modern conda-build to get the conda compilers, as I think that is the "right way" and gives us more robust packages.

What I am looking for is a link or comment (documented in the recipe), which communicates why you think these are the right and necessary options.  That will help me today and possibly your future self in 5 years.

jeanconn:
Gotcha.  Sounds good.  Still fiddling with the new compile directive.

pkg_defs/ska3-template/build.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+mkdir ${PREFIX}/bin
```
jeanconn: This should probably be `mkdir -p` on unix.   Do we need ska3-template or a skare launcher on windows?

jeanconn:
I'll pull out the change to ska_builder.py into a new PR .

pkg_defs/ska3-template/build.sh
taldcroft: Not to my knowledge.  If we find a need then we'll address it.

With regards to the `-p`, hasn't this been working?  Don't we assume from the build process that `${PREFIX}` must exist already?

jeanconn: The "mkdir" fix is here because this recipe it wasn't working for me on conda 4.7 .  ${PREFIX} appeared to exist, but 'bin' did not in that case I think.  I was just thinking "-p" might be more future-proof but probably don't have to worry about that

taldcroft: I am just worried that `mkdir -p` can make any arbitrarily path and do unexpected things.  If `$PREFIX` exists then `mkdir $PREFIX/bin` should never fail, no?

jeanconn:
I assume that the https://github.com/sot/skare3/pull/230/commits/ca2ac95b41440cf086cbc541bfec743fc3564d61 changes are now superseded by #291

PR #235, opened 2019-12-18
dpa_check 2.5.0

jeanconn:
Update dpa_check to v2.5.0


jeanconn:
This was approved at LR with a plan to promote after matlab tools are promoted (which will need to go to FDB tomorrow and which will also be delayed until loads are approved).

jeanconn:
Schedule changed, matlab tools approved and the new plan is not to wait until loads are approved.

jeanconn:
Though I still planned to wait for FDB before promoting.

PR #236, opened 2019-12-18
shiny-related ska_builder.py changes

jeanconn:
shiny-related ska_builder.py changes

### DIFF
--- a/ska_builder.py
+++ b/ska_builder.py
@@ -5,6 +5,7 @@ import subprocess
 import git
 import re
 import argparse
+import platform
 
 
 parser = argparse.ArgumentParser(description="Build Ska Conda packages.")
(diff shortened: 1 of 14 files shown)


jeanconn:
@javierggt I think you had a bunch of pep8 fixes and organizational fixes you wanted to make, not sure about the best merge strategy, but these tweaks I'd like are small in scope.

ska_builder.py
```diff
@@ -151,10 +150,13 @@ def build_list_packages():
 def build_all_packages():
     build_list_packages()
 
+with TemporaryDirectory() as tempska:
```
taldcroft: Assuming import-less building with setuptools_scm, is this still required?

taldcroft:
BTW I am assuming the setuptools_scm updates and other style/code cleanup changes can go to `master` not `shiny`.  The `shiny` branch might not get merged for months since we really do need to wait for the chimchim upgrade to make it operational.  (I pinged Mark Z again earlier today).

jeanconn:
I didn't think chimchim upgrade was a hard lien on moving forward with shiny in ska3/flight.  Though of course a good question about how much you want modern stuff vs possible pain of backports.

ska_builder.py
jeanconn: Good point, but I'm not sure how to work with that in general.  This code is forward compatible so it makes sense to me to just merge these updates and cut the SKA handling when we've made the setuptools_scm package updates.

taldcroft: Yes, but the code burns a hole in my soul.  We can leave it for now so things work, but it has to come out eventually.  That's why I want to get SCM versioning deployed to flight.

taldcroft:
Shiny is only CentOS-6+ compatible.  No ugly backports, all modern.

javierggt:
That was also what I thought, that setuptools_scm changes would go to master. I thought the point of shiny was basically moving to modern package versions.

ska_builder.py
javierggt: I don't get what you guys are discussing here.

jeanconn:
I meant pain of backports if we developed  a feature in a promoted shiny ska3/flight env and then you wanted it right away in centos5 matlab land.

taldcroft:
The point is to not PLAN for and worry about contingencies like that.  Just plan for the shiny `shiny` and if something unexpected comes up then we figure out at that time.

jeanconn:
Right, but I was apparently being so optimistic that I thought shiny could go in January and from your `chimchim` feedback I'm getting pushback.  So I was basically figuring HEAD ska3/flight could perhaps ride ahead of chimchim ska3/flight and be totally independent from ska3 matlab centos5 (which we would plan to not update).  But I guess we'll collect this on the update doc.

taldcroft:
I don't foresee letting GRETA and HEAD diverge significantly for ska3/flight.

PR #237, opened 2019-12-18
Shiny Ska3 (Python 3.8, Numpy 1.18, latest of everything)

jeanconn:
The Ska3 runtime environment consists of an integrated suite of Python-3 packages and
associated applications. The `shiny` distribution is a major update to Ska3 that includes:

- Python 3.6 => 3.8: many very useful new features ([What's New 3.7](https://docs.python.org/3/whatsnew/3.7.html), [What's New 3.8](https://docs.python.org/3/whatsnew/3.8.html)).  Highlights:
  - `dict` is now ordered by insertion order
  - [dataclasses](https://docs.python.org/3/library/dataclasses.html#module-dataclasses)
  - ["walrus" operator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions)
  - [very handy debug print format](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging)
   
- Latest release of every core dependency, often including several years of new development / bug fixes
  - `Numpy` 1.12 (Jan-2017) => 1.18: stricter in many ways for more reliable code but may cause breakage.
  - `Matplotlib` 2.0 (Jan-2017) => 3.2
  - `Scipy` 0.19 (Mar-2017) => 1.5
- Adds [Jupyterlab](https://jupyterlab.readthedocs.io/en/stable/): next generation Jupyter notebook environment
- Update `Chandra.Time` 3.20 =>  4.0 with change that `2020:001` => `2020:001:00:00:00` instead of `2020:001:12:00:00`. This can break code (especially tests) that use the shorthand YYYY:DDD notation for dates.
- Includes `CxoTime` which is now ready for production use.

In addition to distributions built for linux (CentOS-7+) and MacOSX (Catalina), `shiny` adds a fully-functional
and tested binary conda distribution for Windows 10. This will be the basis for FOT MATLAB tools in a future release.


ska_builder.py
```diff
@@ -28,14 +35,14 @@
...
 # And any that are just whitespace
-BUILD_LIST = [b for b in BUILD_LIST if not re.match("^\s*$", b)]
+BUILD_LIST = [b for b in BUILD_LIST if not re.match(r"^\s*$", b)]
 
-if os.uname().sysname == "Darwin":
+if platform.uname().system == "Darwin":
     os.environ["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
 
-if os.uname().machine == 'i686':
+if platform.uname().machine == 'i686':
```
taldcroft: Shiny branch doesn't build on 32-bit, we can cut all this right?

ska_builder.py
```diff
@@ -102,16 +109,25 @@ def get_repo(name, tag):
 def build_package(name):
     print("Building package %s." % name)
     pkg_path = os.path.join(pkg_defs_path, name)
-    cmd_list = ["conda", "build", pkg_path, "--croot",
-                BUILD_DIR, "--no-test", "--old-build-string",
-                "--no-anaconda-upload", "--skip-existing",
+    cmd_list = ["conda", "build", pkg_path,
+                "--croot", BUILD_DIR,
+                "--old-build-string",
+                "--python", "3.6",
```
taldcroft: 3.7?

pkg_defs/xija/meta.yaml
```diff
@@ -16,12 +15,10 @@ requirements:
   # Packages required to build the package. python and numpy must be
   # listed explicitly if they are required.
   build:
-    - pip
     - python
-    - setuptools
     - six
     - numba
-    - numpy >=1.12.1
+    - numpy =1.12
```
taldcroft: For `shiny` I'm guessing we don't need to pin anything, that the top-level numpy directive specifies the build numpy and ska3-flight specifies the runtime?

ska_builder.py
jeanconn: Sure, though you are re-commenting on your own change here.  I don't think we actually need to review this PR.

jeanconn: Though can review if we want any ska_builder improvements in master I suppose.

ska_builder.py
jeanconn: See #236 though hoping to rebuild at 3.8 as the packages become available

taldcroft:
Some current issues with Python 3.8:

- [ ] Sherpa is not yet passing tests.  There are efforts on that path but not clear when it will happen.  Using the latest available it did pass tests with `find_attitude`, but I got a seg fault using the latest `xija_gui_fit`
- [ ] The `rope` package that handles refactoring in VS code isn't working yet on 3.8.
- [ ] Django and deps not in pkgs/main for 3.8

So we might need to settle for 3.7, which seems to have full support and fewer gotchas.  Nevertheless if it ends up being possible to go straight to 3.8 then that would be preferable.

jeanconn:
Regarding outstanding items, I don't quite understand the test fail on Ska.Numpy and if it is a numpy issue or an interface issue from Ska.Numpy.

```
In [5]: Ska.Numpy.test()                                                                                    
=========================================== test session starts ============================================
platform linux -- Python 3.8.2, pytest-5.4.1, py-1.8.1, pluggy-0.13.1
rootdir: /fido.real/shiny/envs/test/lib/python3.8/site-packages
collected 6 items                                                                                          

Ska/Numpy/tests/test_numpy.py ....F.                                                                 [100%]

================================================= FAILURES =================================================
_________________________________________ test_search_both_sorted __________________________________________

start = 1, stop = 10, num = 1000000.0, endpoint = True, retstep = False, dtype = None, axis = 0

    @array_function_dispatch(_linspace_dispatcher)
    def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
                 axis=0):

...

        try:
>           num = operator.index(num)
E           TypeError: 'float' object cannot be interpreted as an integer

numpy/core/function_base.py:117: TypeError

During handling of the above exception, another exception occurred:

    def test_search_both_sorted():
>       a = np.linspace(1, 10, 1e6)

Ska/Numpy/tests/test_numpy.py:47: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
<__array_function__ internals>:5: in linspace
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
```

jzuhone:
@jeanconn you can't have the `num` argument in `numpy.linspace` be a float anymore. The `1e6` is not valid.

jeanconn:
Thanks @jzuhone !

PR #248, opened 2020-01-07
WIP: Add skare3 recipes to build django 3.0.1 and deps

jeanconn:
Add skare3 recipes to build django 3.0.1 and deps

### DIFF
--- a/pkg_defs/asgiref/meta.yaml
+++ b/pkg_defs/asgiref/meta.yaml
@@ -1,0 +1,40 @@
+# Recipe borrowed from https://github.com/conda-forge/asgiref-feedstock
+
+{% set name = "asgiref" %}
+{% set version = "3.2.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 7e06d934a7718bf3975acbf87780ba678957b87c7adc056f13b6215d610695a0
+
+build:
+  noarch: python
+  number: 0
+  script: python setup.py install
+
+requirements:
+  host:
+    - python
+  run:
+    - python
+
+test:
+  imports:
+    - asgiref
+
+about:
+  home: https://github.com/django/asgiref
+  license: BSD-3-Clause
+  license_family: BSD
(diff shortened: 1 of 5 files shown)


jeanconn:
I think just building our own is less problematic than trying to use conda package made with more modern conda versions.

taldcroft:
This is just killing me.  Can you indulge me and show what happens otherwise?

jeanconn:
What is the alternate path?  Grab the sqlparse, asgiref, django from modern repository?  I am concerned that any path isn't the same as your pip install which was what was tested.

taldcroft:
Yes, try grabbing those packages from modern repo and see if it works.  I don't know what you mean by "any path".  These are all three pure Python packages, so it seems quite possible things will just work??

jeanconn:
By "any path" I meant any non-pip way of getting these files into our environment.

And sure, I think I had previously run into the issue that modern packages didn't work for our conda at all (and you had those examples of weird checksum fails) so I figured these borrowed recipes would actually save some time, but fine to try.

jeanconn:
Do you have recommended ways to test out the django installation besides running kadi?

jeanconn:
It didn't die on installation and seems to import, so maybe you are right and the conda-forge linux-64 py36_0 package I grabbed won't be a disaster.  🤞

taldcroft:
For django testing it appears there is no `django.test()`, so testing with kadi is the only thing I can think of.  In practice it is also the only thing that really matters.

jeanconn:
Thanks!, and yes, I'd looked for django.test and any django internal recommendations and didn't see them.  It looks like it is working.  With regard to the "only thing that matters" I figured that the web server would use the same conda package, but if we run into a new fail in that testing we can address then.

I'm not sure about the best place to capture/document that the asgiref, sqlparse, and django conda packages are the ones that I retrieved from conda-forge.

jeanconn:
@taldcroft Oops.  I forgot to go and check this for 32bit.   I can't find built packages at the right version (I know you keep saying Django isn't really a binary package but not sure if we want to pip install in the ska3/matlab conda environment or make our own conda package.)

taldcroft:
Yes, I don't understand why django packages are separated by arch on conda but not on PyPI.  Anyway, looks like there is no choice but rolling our own, again..  Sigh.

PR #285, opened 2020-01-17
Skare3 tools

javierggt:
This is still work in progress. Here so you can see it.

### DIFF
--- a/skare3/dashboard/ska_github_info.py
+++ b/skare3/dashboard/ska_github_info.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+
+from skare3.tools.github import GithubAPI
+import re
+import json
+import argparse
+
+
+API = GithubAPI()
+
+
+def get_repository_info(owner_repo):
+    global API
+    owner, repo = owner_repo.split('/')
+
+    last_tag = API.get(f'repos/{owner}/{repo}/releases/latest').json()
+    tag_info = API.get(f'repos/{owner}/{repo}/git/ref/tags/{last_tag["tag_name"]}').json()
+    tag_sha = tag_info['object']['sha']
+    rel_commit = API.get(f'repos/{owner}/{repo}/commits/{tag_sha}').json()
+    commit_date = rel_commit['commit']['author']['date']
+
+    commits = API.get(f'repos/{owner}/{repo}/commits',
+                      params={'sha':'master', 'since': commit_date}).json()
+    commits = commits[:-1]  # remove the commit associated to the release
+
+    branches = API.get(f'repos/{owner}/{repo}/branches').json()
+    n_branches = len(branches)
+
+    issue_page = API.get(f'repos/{owner}/{repo}/issues', params={'per_page':100}).json()
+    issues = issue_page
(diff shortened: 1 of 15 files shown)


jeanconn:
Looks great so far, thanks for the work on this!

skare3/dashboard/ska_github_info.py
```diff
@@ -0,0 +1,117 @@
...
+        'sot/parse_cm',
+        'sot/proseco',
+        'sot/pyyaks',
+        'sot/ska_path',
+        'sot/ska_sync',
+        'sot/sparkles',
+        'sot/starcheck',
+        'sot/tables3_api',
+        'sot/testr',
+        'sot/xija'
```
jeanconn: I think this is, offhand, missing find_attitude and ska_helpers to start.  Not sure if other not-ska3-flight not-Python packages like task_schedule and watch_cron_logs belong here too.  

And not sure if having the list hard-coded here is the best way to manage it?  But probably sufficient given how we develop packages.

taldcroft: Seems like a dynamically-acquired list of every repo in the sot org might be the way to go.  One can apply filters, e.g. for a repo that has had no commits in the last N years, or whatever.  But in general I'm interested in seeing activity and status across the board.  This can point to repos that might be best "archived", who knows.

taldcroft: Of course it might turn out that we lower the S/N of the table too much by having a bunch of inactive repos.  Not sure offhand.

jeanconn: I was wondering if something like a Watched label could be used on the github side to make the list.

javierggt: We can have everything!

One could get all repositories of a given "owner" (in our case 'sot' or 'acis').

javierggt: that's the point of having filters on top. If the filters do not work for us, we make the list smaller.

taldcroft: I like the owner idea.  In fact I bet @jzuhone would be happy to get this status dashboard for `acisops`.  (WIP at http://cxc.cfa.harvard.edu/mta/ASPECT/skare3/dashboard/)

taldcroft:
Now that I've posted the link on a public page (...) it occurs to me to ask whether there are any security concerns.  This page does take inputs and run some javascript for the table, so maybe it needs to go back to icxc?  Is the table javascript library considered secure and maintained?

skare3/dashboard/ska_github_info.py
taldcroft: I think that filtering repos by age of last commit (which could be a new column in the table, in years) would be reasonable.  I'm not aware of any way to label a repository as a whole.  If it does exist, I would have the label opt out of being watched rather than opt in.

javierggt:
good question... I do not think it is maintained.

taldcroft:
Right, so with apologies for the churn, but probably best to move it back to icxc.

javierggt:
This lives in its own repository now.

PR #291, opened 2020-02-01
Updates to shiny branch for Python 3.8

### LINKED ISSUE #289: Packages without recipe build dependences on setuptools_scm and git_archive will fail?
For many of the packages, the meta.yaml in the package recipe does not include a setuptools_scm and setuptools_scm_git_archive dependence in the build dependencies.  ska_helpers has a setuptools_scm run dependence but not build dependence.

What this means, is that if setuptools_scm and setuptools_git_archive are not in the ska3build environment, they are pip downloaded as eggs in to the individual package repo.  They are then found and used when conda does the package build if needed.

If setuptools_scm and setutools_git_archive are in the ska3build environment, they are not pip downloaded and the conda build fails because they are also not in the build meta.yaml, so you just end up with unsatisfied dependencies.

I think the answer is just to add setuptools_git_archive to the ska_helpers run dependencies and then add ska_helpers to the build dependencies of all of the modules.  But then the question is if we should set the build dependencies in the meta.yaml by some environment variable or other process instead of doing this one edit in all the files this time.

taldcroft:
This includes:

- Updates for Python 3.8 support
- Bulk updates to recipes to make minimal build requirements
- Remove packages we are no longer building locally like astropy
- Other simplifications where possible

Known issues / highlights:
- sherpa conda build not yet available for Py3.8 and apparently not passing tests yet? 
 See https://github.com/sherpa/sherpa/pull/696.  Should be able to build ourselves if needed.
- django 3.0 is an ots/ install from conda-forge
- Tested / built on MacOSX only
- kadi recipe does not include django as a dependence.  This did not work the last time I tried.
- Includes new `ska3-flight-latest` and `ska3-core-latest` metapackages.  These are intended to replace `ska3-dev`, with the feature of being separable.  E.g. you can install only `ska3-flight-latest` and see the minimal requirements.  Depending on channel this might be the latest *releases* or latest *masters*.

Fixes #289
Closes #230 
Closes #236


taldcroft:
I think mica depends on tables3_api still, though I'm not sure if that is only in processing code.

ska_builder.py
```diff
@@ -40,14 +40,7 @@
...
 if platform.uname().system == "Darwin":
-    os.environ["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
-
-if platform.uname().machine == 'i686':
-    # Skip starcheck and ska3-perl on 32 bit
-    for pkg in ['starcheck', 'ska3-perl']:
-        if pkg in BUILD_LIST:
-            BUILD_LIST.remove(pkg)
-
+    os.environ["MACOSX_DEPLOYMENT_TARGET"] = "10.14"  # Mojave
```
jeanconn: I think this wasn't really working in the old version of conda anyway (not really setting to 10.9).  If the variable works now, will probably need to scrub the twiki docs for all references of machine `zero` as I don't think the imac supports mojave.

jeanconn:
Thanks for the work on this!  Though I'm not sure about the best way to manage shiny vs master at this point (with regard to things like handling #289) and when we might go live with shiny.

ska_builder.py
jeanconn: Granted, if @javierggt 's OSX building is ready-to-go all the `zero` stuff is probably moot anyway.  Just don't have good ways to capture changes to docs in the accompanying code with use of the wiki.

taldcroft:
I think this closes #289, so that is not an issue.  Otherwise I don't see any reason to get stalled on this because of uncertainties in the details.  We'll go live with shiny at some point when it is ready, which won't happen until we just dive in and start trying to really make it work.  And right now I am scratching that itch!!

ska_builder.py
taldcroft: Why was this in there to begin with?  Can we just cut it out?

taldcroft:
One lesson-learned and something I'll be applying is to not include secondary dependencies.  For instance `Ska.ftp` had a run dependence on `pycrypto`, but this is never imported anywhere in our code.  Instead that presumably was a dependence of `paramiko`, but that is no longer a dependency.  But before discovering that I ended up spinning around because `pycrypto` is no longer supported in `defaults`, and I found it in conda-forge, but that's not great.

So for instance I'll be removing `hdf5` and (various other packages that we don't import directly) from `ska3-dev` (which is turning into `ska3-flight-latest` and losing the `ska3-core` bit).  Of course for `ska3-flight` we will still want to pin everything per current practice.

ska_builder.py
taldcroft: The plan with build automation is that the wiki gets chopped down in a big way.  But if you wanted to, it is easy enough to branch the current maintainers wiki into a shiny version and start editing that to reflect how things will be post-shiny.

taldcroft:
The conda dependency solver is beyond me somehow.  As another example I tried to install `annie` and it fails to resolve the dependency graph even though every other dependent package is already installed!  So I remove `llvmlite` from the `run` dependencies since what annie really needs is `numba`.  Now it solves the graph immediately.

Another example.  I try installing kadi with a recipe that includes django as a run dependency.  Since django 3.0+ is not in defaults yet, I downloaded 3 packages from conda-forge and put them into an `ots` directory:
```
conda install -c file://$PWD/ots -c file://$PWD/builds -c defaults kadi
# FAIL
```
Instead, take django out of the kadi dependencies, waste a long time rediscovering caching of the build package in a hidden `.cache` dir, and then do the following:
```
conda install -c file://$PWD/ots -c file://$PWD/builds -c defaults kadi
conda install -c file://$PWD/ots -c file://$PWD/builds -c defaults django
# SUCCESS
```

taldcroft:
This is finally working for me to give a build for all packages except the ACIS ones.  It installs and passes at least kadi and cheta unit tests.  I haven't tried more.  (All statements apply only to MacOS Catalina).

I'd like to merge this tomorrow to get it out of my system.  It's just too big to keep twiddling, so we can have post-merge discussions and fixes / changes as needed.

jeanconn:
What does "django 3.0 is an ots/ install from conda-forge" mean as an issue?

jeanconn:
Regarding " Otherwise I don't see any reason to get stalled on this because of uncertainties in the details. "  I wasn't concerned about shiny being stalled (were we stalled?) I just want some of these changes in master now to keep building working; but maybe I won't need to build all the packages again (ever!).

Also, regarding the solver hiccups, I'd assume you'd need  "--strict-channel-priority" to just lead with the "ots" like you seem to want it, but not sure what you've got in those build directories to know.  I also don't know what "not including a secondary dependence" really means in this context... is that a redundant explicit dependence?

pkg_defs/Chandra.Maneuver/meta.yaml
```diff
@@ -40,4 +42,4 @@ test:
 
 
 about:
-  home: git@github.com:sot/Chandra.Maneuver
+  home: https://github.com/sot/Chandra.Maneuver
```
jeanconn: This looks fine, though I think I left some of these with ssh style because the auth worked better for me.  Will see if that just needs auth fixes for the aca user.

javierggt: That's a good question... the automated build uses http because it is easier to authenticate. If you check, I jumped through some hoops to do that. I personally prefer ssh.

taldcroft: Ah, that's because the `about: home` tag is being (mis-)used as the clone URL instead of the intent of being the project home.  We might consider populating the `about: dev_url` tag instead which is closer to the idea of a clone URL. 

The curious thing is that this does work for me on my mac, and I have 2FA so generally speaking https has presented problems (requiring a token).  I'm not exactly sure what is happening under the hood to allow this https: access to successfully clone a repo.  I just tested `git clone https://github.com/sot/maude.git` for aldcroft@mac (works) and aca@kady (fails). 

Anyway I'll just revert this to reduce problems.

taldcroft:
>  I'd assume you'd need "--strict-channel-priority" to just lead with the "ots" like you seem to want it, but not sure what you've got in those build directories to know. 

Indeed the information about it is in my build directories is relevant:
```
(shiny38) ➜  skare3 git:(shinier) ✗ find ots -name '*.tar.bz2'
ots/osx-64/django-3.0.2-py38_0.tar.bz2
ots/osx-64/nb_conda_kernels-2.2.2-py38_0.tar.bz2
ots/osx-64/nb_conda-2.2.1-py38_2.tar.bz2
ots/noarch/sqlparse-0.3.0-py_0.tar.bz2
ots/noarch/asgiref-3.2.3-py_0.tar.bz2
```

So basically just django and deps and nb_conda.  The upshot there is that I don't believe that `--strict-channel-priority` would make any difference in most of the problems I was having because those packages were not always players.  BUT, I could be wrong since conda rarely gave me anything useful to work with to debug.

taldcroft:
I reverted the URLs for cloning.

PR #298, opened 2020-02-10
git clone with https instead of ssh

javierggt:
Some of the red in the dashboard comes from this.

Automated build uses only basic authentication over https. I did not include ssh keys for the build process. For this reason, I would replace all uses of git@github.com: for the https url.

There are other possible solutions, I just chose the simplest one:
- add ssh keys
- make it an option of the script.

### DIFF
--- a/ska_builder.py
+++ b/ska_builder.py
@@ -21,6 +21,8 @@ parser.add_argument("--build-list", default="./ska3_flight_build_order.txt",
                     help="List of packages to build (in order)")
 parser.add_argument("--force", action="store_true",
                     help="Ignore already built packages, build them again.")
+parser.add_argument("--github-https", action="store_true", default=False,
+                    help="Authenticate using basic auth and https. Default is ssh.")
 
 args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


javierggt:
I just added an option to either use https or ssh. It either uses ssh or https on all packages (right now some packages have https or ssh, and one usually has a preferred authentication method).

javierggt:
Is that an official approval? :-)

I did test that by default annie/Quaternion was fetched using ssh and with the option it used https. They are examples of meta.yaml hacing ssh/https respectively.

jeanconn:
Eh.  I hit the "approve" button when I commented, so that's official.  Overall I don't think we need a super-high bar on changes to ska_builder.py as there aren't really credible paths by which we'll build a package "incorrectly" with these kinds of changes.

taldcroft:
BTW, my current practice is to delete PR branches immediately upon merge to keep things tidy.  Is there a reason to not do that?

javierggt:
no, no reason. It should be deleted.

PR #309, opened 2020-02-20
Add skare3 recipe to build django 3.0.1 

jeanconn:
I'm not seeing a way around this just now.  Will brainstorm.

I couldn't find a repo in which somebody was actually building modern packages like django 3.0.1 on 32 bit.


jeanconn:
Seems to be working after https://github.com/sot/skare3/pull/309/commits/4a4b147292433810170c601bd0b1c12e23aa3ae0

jeanconn:
We probably don't need the patches (haven't tested with or without) but don't care.

jeanconn:
I was planning to only use the package made with this on linux-32.  I don't think we have a driver to build/replace our working linux-64 one.

PR #321, opened 2020-04-06
Add aca view to skare3

javierggt:
This PR does not add aca_view to ska3-flight. I think that is ok. One can still do `conda install aca_view`. Would you prefer to add it to ska3-flight?

### DIFF
--- a/pkg_defs/aca_view/build.sh
+++ b/pkg_defs/aca_view/build.sh
@@ -1,0 +1,1 @@
+pip install --no-deps --verbose --no-binary :all: --no-index .
--- a/pkg_defs/aca_view/meta.yaml
+++ b/pkg_defs/aca_view/meta.yaml
@@ -1,0 +1,50 @@
+package:
+  name: aca_view
+  version:  {{ SKA_PKG_VERSION }}
+
+build:
+  noarch: python
+  script_env:
+    - SKA_TOP_SRC_DIR
+
+source:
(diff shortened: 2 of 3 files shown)


javierggt:
We need to close this before closing sot/skare3/pull/342

pkg_defs/aca_view/meta.yaml
```diff
@@ -0,0 +1,50 @@
+package:
+  name: aca_view
+  version:  {{ SKA_PKG_VERSION }}
+
+build:
+  noarch: python
+  script_env:
+    - SKA_TOP_SRC_DIR
```
jeanconn, line 8: Probably not needed, but if you've built the package with this for 2020.4rc1 can leave it?

pkg_defs/aca_view/build.sh
```diff
@@ -0,0 +1 @@
+pip install --no-deps --verbose --no-binary :all: --no-index .
```
jeanconn, line 1: And this could also be python setup.py install etc but if already built this is also fine.  Not sure where we should take another action to do more recipe cleanup or if we'll just do that for shiny?

javierggt: Right. I guess these changes can be made for shiny (that was my assumption when John asked about the acis packages the other PR). Either that, or we create an issue about it now.

PR #323, opened 2020-04-10
Add fep1_actel_check, fep1_mong_check, bep_pcb_check to skare3

jzuhone:
This PR adds the following packages to skare3:

http://github.com/acisops/fep1_actel_check
http://github.com/acisops/fep1_mong_check
http://github.com/acisops/fep1_actel_check

which are getting upgrades in conjunction with acis_thermal_check v3.0.

I would like to have them in ska3-flight, but we can decide when that should happen.

### DIFF
--- a/pkg_defs/bep_pcb_check/build.sh
+++ b/pkg_defs/bep_pcb_check/build.sh
@@ -1,0 +1,2 @@
+export SKA=/proj/sot/ska
+pip install --no-deps --verbose --no-binary :all: --no-index .
--- a/pkg_defs/bep_pcb_check/meta.yaml
+++ b/pkg_defs/bep_pcb_check/meta.yaml
@@ -1,0 +1,44 @@
+package:
+  name: bep_pcb_check
+  version:  {{ SKA_PKG_VERSION }}
+
+build:
+  noarch: python
+  script_env:
+    - SKA_TOP_SRC_DIR
+
+source:
+  path: {{ SKA_TOP_SRC_DIR }}/bep_pcb_check
+
+
+# the build and runtime requirements. Dependencies of these requirements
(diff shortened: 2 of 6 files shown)


pkg_defs/bep_pcb_check/build.sh
```diff
@@ -0,0 +1,2 @@
+export SKA=/proj/sot/ska
+pip install --no-deps --verbose --no-binary :all: --no-index .
```
jeanconn, line 2: We're pushing to update the installs to use

```
python setup.py install --single-version-externally-managed --record=record.txt
```
Will probably move those into the meta.yaml at some point too, but at this point it would probably be good just to get the `python setup.py` transition.  Probably also don't need to define SKA.

pkg_defs/bep_pcb_check/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  # Packages required to build the package. python and numpy must be
+  # listed explicitly if they are required.
+  build:
+    - pip
+    - python
+    - setuptools
+    - six
+    - numpy
+    - matplotlib
+    - acis_thermal_check
```
jeanconn, line 26: And I *think* everything now needs setuptools_scm in build.  Right @javierggt ?

jeanconn: Oops, it looks like these new packages aren't set for setuptools_scm / ska_helpers so that previous comment may be moot for now?

jzuhone:
@jeanconn I copied the `meta.yaml` and `build.sh` files for these from the `dpa_check` one, so if we need to fix these then we need to fix the other ACIS package configurations. Is that what we should do here too?

PR #326, opened 2020-04-10
Add conda recipe for skare3 tools

javierggt:
This recipe ends up pulling several packages that are not currently in our repository. That is thanks to these three dependencies:
- google-auth-oauthlib
- google-api-python-client
- sphinx-argparse
The last one is just for documentation (useful for documenting scripts).

Where this goes is an open question. At least I wanted to have it somewhere documented.

### DIFF
--- a/pkg_defs/skare3_tools/build.sh
+++ b/pkg_defs/skare3_tools/build.sh
@@ -1,1 +1,0 @@
-python setup.py install --single-version-externally-managed --record=record.txt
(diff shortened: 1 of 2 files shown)


jeanconn:
I think this is fine, but I'm concerned about pulling in anything new into master if we still plan on another Matlab / CentOS 5  release before the centos7/8 update.

taldcroft:
Since this PR doesn't actually add the package into ska3-flight there should be no concern about dependencies and CentOS-5.

However, please remove `build.sh` and put the build command into `meta.yaml`. With that I'm good to merge this.

PR #329, opened 2020-04-11
No eggs

javierggt:
This PR changes the build of namespace packages to not use eggs.

While doing that, I also replace all uses of `pip` by `python setup.py` and this required explicitly adding setuptools_scm to requirements.

And while doing that... I went and did it also for non-namespace packages.

Now... will this be a testing nightmare?

### DIFF
--- a/pkg_defs/Chandra.Maneuver/build.sh
+++ b/pkg_defs/Chandra.Maneuver/build.sh
@@ -1,1 +1,1 @@
-pip install --no-deps --verbose --no-binary :all: --no-index --egg .
+python setup.py install --single-version-externally-managed --record=record.txt
--- a/pkg_defs/find_attitude/meta.yaml
+++ b/pkg_defs/find_attitude/meta.yaml
@@ -33,7 +33,7 @@ requirements:
     - quaternion
     - ska.quatutil
     - sherpa
-    - ska_helpesr
+    - ska_helpers
 
 test:
   imports:
--- a/pkg_defs/find_attitude/meta.yaml
+++ b/pkg_defs/find_attitude/meta.yaml
@@ -18,7 +18,8 @@ requirements:
   # listed explicitly if they are required.
   build:
     - python
-    - pip
+    - setuptools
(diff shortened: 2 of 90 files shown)


pkg_defs/Chandra.Maneuver/build.sh
```diff
@@ -1 +1 @@
-pip install --no-deps --verbose --no-binary :all: --no-index --egg .
+python setup.py install --single-version-externally-managed --record=record.txt
```
jeanconn, line 1: I don't care if these changes are in build.sh or if the script is moved to meta.yaml but wasn't sure which way we wanted to go for this.

pkg_defs/find_attitude/meta.yaml
```diff
@@ -32,7 +33,7 @@ requirements:
     - quaternion
     - ska.quatutil
     - sherpa
-    - ska_helpers
+    - ska_helpesr
```
jeanconn, line 36: That one can't be right.

jeanconn:
I think I'm fine with these, but I was thinking they'd all go in the 2020.3 PR.  Either way.

pkg_defs/find_attitude/meta.yaml
javierggt: oops

javierggt: Good catch. Fixed

pkg_defs/Chandra.Maneuver/build.sh
javierggt: Also here I do not understand exactly what you mean. I made the change to be minimally invasive, but if you prefer another way, we can do that.

taldcroft: Eventually all the `build.sh` are going to be deleted as in shiny and use `script:` in `meta.yaml`.  This is the easier way to do for also supporting Windows builds.  But the changes here are fine now for getting this release out.

PR #340, opened 2020-04-20
added new acis packages bep_pcb_check fep1_actel_check fep1_mong_check

javierggt:
Changes are still not finished.

### DIFF
--- a/pkg_defs/bep_pcb_check/meta.yaml
+++ b/pkg_defs/bep_pcb_check/meta.yaml
@@ -4,8 +4,6 @@ package:
 
 build:
   noarch: python
-  script_env:
-    - SKA_TOP_SRC_DIR
 
 source:
   path: {{ SKA_TOP_SRC_DIR }}/bep_pcb_check
--- a/pkg_defs/bep_pcb_check/meta.yaml
+++ b/pkg_defs/bep_pcb_check/meta.yaml
@@ -1,0 +1,40 @@
+package:
+  name: bep_pcb_check
+  version:  {{ SKA_PKG_VERSION }}
+
+build:
+  noarch: python
+  script_env:
+    - SKA_TOP_SRC_DIR
+
+source:
+  path: {{ SKA_TOP_SRC_DIR }}/bep_pcb_check
+
+
+# the build and runtime requirements. Dependencies of these requirements
+# are included automatically.
+requirements:
+  # Packages required to build the package. python and numpy must be
+  # listed explicitly if they are required.
+  build:
+    - python
+    - setuptools
+    - setuptools_scm
+
+  # Packages required to run the package. These are the dependencies that
+  # will be installed automatically whenever the package is installed.
+  run:
+    - python
+    - matplotlib
+    - dpa_check
+    - acis_thermal_check
+    - ska_helpers
+    - testr
+
+test:
+  imports:
+    - bep_pcb_check
+
+about:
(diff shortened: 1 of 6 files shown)


jzuhone:
I still need some clarification from @jeanconn as to whether or not we need to edit the other ACIS packages similarly.

pkg_defs/bep_pcb_check/meta.yaml
```diff
@@ -0,0 +1,40 @@
+package:
+  name: bep_pcb_check
+  version:  {{ SKA_PKG_VERSION }}
+
+build:
+  noarch: python
+  script_env:
+    - SKA_TOP_SRC_DIR
```
jeanconn, line 8: I think this is not required.

jeanconn:
I'm fine with that being up to @javierggt .  I think we're just talking about recipe improvements (so probably don't *need* the updates but not sure when it is most convenient to get them in).

jeanconn:
And @javierggt did you just make this to supersede #323 ?

javierggt:
Oops, no, I did not intend to superseed that one. Sorry about that. I think I will just close this one.

jeanconn:
The tiny issue is that we can only make suggestions to John's PR.  So if you just want to get it in with these bits that use `python setup.py install` that's fine.

javierggt:
I wonder if the dependencies in this PR are better. I created them by using findimports. And the build dependencies should be a rather short list.

pkg_defs/bep_pcb_check/meta.yaml
javierggt: Right. I removed them.

javierggt:
So... I decided to go with this one in the end. The main difference is the requirements, I believe.

I am listing only explicit requirements:
- numpy is not a requirement because it is not imported explicitly. matplotlib is imported explicitly.
- six is not imported by any of these
- added ska_helpers and testr to all because they are imported.

jeanconn:
I'm fine with only explicit requirements.  In the end it doesn't buy much (these require acis_thermal_check which has numpy as dependence) but doesn't cause any problems.

jzuhone:
@javierggt happy for this to be the PR. My only question is if the other ACIS packages (`dpa_check`, `dea_check`, `psmc_check`,`acisfp_check`) should have a similar setup, because right now they do not.

javierggt:
@jzuhone, I think it is not necessary at this time (for this coming release). In the future, that will certainly be done.

javierggt:
We need to close this before closing sot/skare3/pull/342

jeanconn:
@jzuhone I've just got really simple run tests of the other checkers in our test suite.  Did you want to suggest similar run/smoke tests for us to run for the new checkers?

jzuhone:
@jeanconn did you want to try the new answer testing? it should just run with e.g.:

```python
import psmc_check
psmc_check.test()
```

jeanconn:
Yes!  I think we probably want both?  In that we have the "run like they do for load review" stuff like

https://github.com/sot/ska_testr/blob/8092f9a5a7ad2036fcc962006602def7796f12da/packages/acisfp_check/test_regress.sh

but should also add the new answer testing?  Or do you think the new answer testing trumps the "run like they do for load review" tests?  Doesn't belong in this PR anyway but...

jzuhone:
Probably do both, to be honest.

PR #380, opened 2020-05-27
add sync-conda step to pre-release flow

javierggt:


taldcroft:
What's happening with this one?  I don't really care personally, but good to merge sooner rather than later if it is wanted.

javierggt:
These changes are not needed after other PRs.

PR #389, opened 2020-07-02
Allow changing the org of a repo and improve --force

taldcroft:
This allows changing the org of a repo, e.g. from acisops to sot in the case of a forked repo.

This PR is conservative and removes the repo if you are playing around with the org to ensure that upstream and origin are always correct.  For the purpose of code simplicity it always makes a new remote `upstream` which is where tags come from, but `upstream` is usually the same as `origin`.

I also made it so that when `--force` is supplied that the cache files and cached package are removed. I've noticed some problem with the cache files being there and processing not happening like I expected. I never totally understood what was happening but eventually was doing this cache cleaning by hand when rebuilding the same package.

## Testing

- [x] Functional testing.

I ran this for `backstop_history` in the acisops org and successfully built the `shiny` branch which only exists in `sot`.  This has the git@github style of URL.  

I also ran for `kadi` setting the org to `taldcroft`.  This has the `https:github` style.  It failed to find the repo in taldcroft but I confirmed the regex substitution worked.

I also ran with --force and confirmed that the expected files were removed.

### DIFF
--- a/ska_builder.py
+++ b/ska_builder.py
@@ -6,6 +6,7 @@ import git
 import re
 import argparse
 import platform
+import shutil
 
 
 parser = argparse.ArgumentParser(description="Build Ska Conda packages.")
@@ -37,6 +38,9 @@ parser.add_argument("--numpy",
                     help="Build version of NumPy")
 parser.add_argument("--github-https", action="store_true", default=False,
                     help="Authenticate using basic auth and https. Default is ssh.")
+parser.add_argument("--github-org",
+                    help="Use this org instead of org in meta (mostly for forked packages)")
+
 
 args = parser.parse_args()
 
@@ -57,9 +61,15 @@ BUILD_DIR = os.path.abspath(os.path.join(args.build_root, "builds"))
 SRC_DIR = os.path.abspath(os.path.join(args.build_root, "src"))
 os.environ["SKA_TOP_SRC_DIR"] = SRC_DIR
 
+
 def clone_repo(name, tag=None):
     print("  - Cloning or updating source source %s." % name)
     clone_path = os.path.join(SRC_DIR, name)
+
+    if args.github_org and os.path.exists(clone_path):
+        print('Removing git clone because --github-org was supplied')
+        shutil.rmtree(clone_path)
+
     if not os.path.exists(clone_path):
         metayml = os.path.join(pkg_defs_path, name, "meta.yaml")
         meta = open(metayml).read()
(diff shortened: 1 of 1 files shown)


ska_builder.py
```diff
@@ -122,8 +147,26 @@ def build_package(name):
...
+
+    if args.force:
+        for path in Path(BUILD_DIR).glob(f'*/.cache/*/{name}-*'):
+            print(f'Removing {path}')
+            path.unlink()
+
+        sys_prefix = Path(sys.prefix)
+        if (sys_prefix.parent).name == 'envs':
+            # Building in a miniconda env, can find packages one dir up in pkgs
+            pkgs_dir = sys_prefix.parent.parent / 'pkgs'
```
javierggt, line 159: I don't understand this. If I am within an environment and I run from the source skare3 directory, this will still set pkgs_dir to '/Users/javierg/SAO/miniconda3/pkgs' and delete all directories named like the package in there. Is that right? Why would conda pull use that directory? Why should this script mess with a directory out of its scope?

ska_builder.py
```diff
@@ -137,6 +180,11 @@ def build_one_package(name, tag=None):
     build_package(name)
     print('')
 
+    if args.github_org:
+        print('Removing git clone because --github-org was supplied')
+        clone_path = os.path.join(SRC_DIR, name)
+        shutil.rmtree(clone_path)
+
```
javierggt, line 187: This is another thing... you are removing the directory if you want to force the build? This does more than forcing the build. The --force used to just make sure conda didn't skip it. This starts fresh.

I did not like much the previous way, where the script could be fooled into building something with extra commits in the local index, and I though we should check for that if we wanted to ensure a clean build.

This ensures the clean build, but only if you force the conda build.

ska_builder.py
taldcroft: This is for the corner case of building and re-building packages at the same version (and uploading and then making a new env for testing from the "official" repo). When you do that there may already be this cached version in `pkgs` and so it won't bother re-downloading it, conda just uses the existing cached version. I've been bit by this several times, along with problems related to caching in `builds`.

javierggt: so maybe adding another option like --clear-cache?

ska_builder.py
taldcroft: This is removing the repo in `src` only when you supply `--github-org` (which is rare). Think about `backstop_history` that comes from `acisops`.  If you have been building that before and have repo in your `src` dir cloned from `acisops`, and then supply `--github-org=sot` then it won't work because `origin` is acisops not sot.  So I decided to be safe to just disable the git repo caching in `src` when you are playing around with the org with `--github-org`.

I don't quite understand your question related to `--force` since that is not related to this code. Even if there is no repo it won't "force" the build if the version is the same, though of course without a repo it will always re-clone the repo.

javierggt: ooops, right. I don't know why I read --force.

ska_builder.py
taldcroft: In principle `--clear-cache` is the right thing, but in practice you need to clear the cache if and only if you are using `--force`. So I don't think the added complexity is winning us anything.

javierggt: You might not be aware of a way we (@jeanconn and I) have used this script on ocasion.

To make a quick test, you can make a change locally, commit (do not push) and build with the --force option. In this case, the current script does not realize that it is being tricked. Your change would prevent this. I think it is convenient to have a mechanism like that to test without having to push upstream.

javierggt: Of course, one could remove the output package from the output directory, reindex, and run without `--force`. But this is exactly doing by hand what `--force` is supposed to do.

taldcroft: The main use case for `--force` is changing `skare3` itself (`ska_builder.py` or a recipe), since if you update the package source and commit, that makes a new version of the package (yay SCM versions!) and force is not needed anyway. So this is why I feel like there is no need for a separate cache clearing option.  I found that telling `conda build` to ignore the existing build was not enough, that I have to remove those cache files in `builds`. 

And then on the testing side if you make a new package at the same version, and then upload that new package to the web server repo, you will not get the new package with `conda create` unless you remove it from the `pkgs` cache.

Conversely, if you have not made a new package build at the same version, I can't see any reason why the cache files need to be cleared.

PR #390, opened 2020-07-04
WIP: refactor, modernize, improve ska_builder.py

taldcroft:
This is a WIP for making `ska_builder.py` a bit more organized, simpler and modern.

A key functional change is defaulting to just using the packages in `pkg_defs`, agreeing now with the suggestion from @javierggt.  Having looked more carefully at the diffs from what is there vs. the build order file, basically we either want to just build the packages (like the new ACIS checkers) or remove the recipe (like jplephem).

It also now allows specifying one or more packages to build by name.

Also removes the caching of git repos entirely in order to make the code simpler and more reliable.  From my experience the time for cloning is a small part of the overall build time.

### DIFF
--- a/pkg_defs/Chandra.Maneuver/meta.yaml
+++ b/pkg_defs/Chandra.Maneuver/meta.yaml
@@ -20,8 +20,6 @@ requirements:
     - setuptools
     - setuptools_scm
     - setuptools_scm_git_archive
-    - quaternion
-    - ska_helpers
   # Packages required to run the package. These are the dependencies that
   # will be installed automatically whenever the package is installed.
   run:
--- a/pkg_defs/backstop_history/meta.yaml
+++ b/pkg_defs/backstop_history/meta.yaml
@@ -20,8 +20,6 @@ requirements:
     - setuptools
     - setuptools_scm
     - setuptools_scm_git_archive
-    # Mistake in backstop_history/setup.py:
-    - ska_helpers
 
   # Packages required to run the package. These are the dependencies that
   # will be installed automatically whenever the package is installed.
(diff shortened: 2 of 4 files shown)


javierggt:
I will check it. My only comment right now is that I would like to be able to test the build without having to push upstream. How would that be if you remove the local caching of git repositories?

javierggt:
I suppose one way is to be able to specify a repository in a local directory.

javierggt:
ah, and another important thing:

I mentioned somewhere else that I intended to wait for shiny and then combine this with what I have in skare3_tools. Specifically, there is [this script that wraps ska_builder.py](https://github.com/sot/skare3_tools/blob/ci-testing/actions/build/files/build.py) to add a few options. We should merge them so there are not two scripts for essentially the same thing.

taldcroft:
I've changed this to use a more generic override of `--repo-url` so you can point to a local repo via `file:///...`.  I think that should more cleanly allow your use case without relying on a persistent clone that is sitting in `src`. This separates the process and directory structure of *building* from the maintenance of a source git repo.

Does that work?

taldcroft:
About your wrapper in skare3_tools, sounds fine but let's try to get shiny going before thinking about merging.

About allowing `--tag` to take a general ref, I agree in principle. If we get rid of the cached repo and always clone, I think this will be trivial.  If you rely on a persistent clone then one needs to do a `git pull` in general to be sure it is up to date, but then `git pull <commit>` doesn't work.  So that's another reason to always clone fresh.

taldcroft:
With 40f963c you can build at a tag specified by any valid git ref, e.g. a commit hash.

PR #394, opened 2020-07-09
Use the anaconda compilers on osx and linux

jeanconn:
Use the anaconda compilers on osx and linux

My understanding is that the anaconda clang on OSX will need access to an SDK and is configured by default to use /opt/MacOSX10.9.sdk.  I haven't figured out how to use a different one or how to get it to respect MACOSX_DEPLOYMENT_TARGET and CONDA_BUILD_SYSROOT . 

It also looks to me like if /opt/MacOSX10.9.sdk isn't present but there is an SDK installed in /Library from Xcode, the anaconda clang seems to work with that without error.

### DIFF
--- a/pkg_defs/Chandra.Time/meta.yaml
+++ b/pkg_defs/Chandra.Time/meta.yaml
@@ -14,6 +14,7 @@ requirements:
   # Packages required to build the package. python and numpy must be
   # listed explicitly if they are required.
   build:
+    - clangxx_osx-64 # [osx]
     - python
     - setuptools
     - setuptools_scm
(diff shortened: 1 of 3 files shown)


taldcroft:
I'm trying this out, but it needs to be rebased on current shiny to work e.g. with `python ska_builder.py --force --tag shiny Chandra.Time`.

taldcroft:
Is there something for Windows?

jeanconn:
Why does it need rebase?

jeanconn:
And as far as I can tell there is nothing for Windows.  https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html

jeanconn:
But it looks like conda-forge may be building packages for windows with 'vs2015', so that might just be more recent than the anaconda 5 compiler notes in the conda-build documentation.  And maybe it would just work with the {{ compiler('c') }} etcsyntax.

taldcroft:
> Why does it need rebase?

It wasn't working with --repo-url.  :frowning:

taldcroft:
On Windows, that's fine, I was just checking for completeness.

PR #405, opened 2020-07-15
Allow for meta.yaml compiler option in ska_builder

### LINKED ISSUE #391: ska_builder.py incompatible with modern compiler recipes
**Description**

It looks like the standard for modern recipes is to use jinja to define the compiler dependencies and that doesn't work with ska_builder.py.

As in, if I try to add 
```
 requirements:
     build:
       - {{ compiler(‘c’) }}
```
to the xija recipe, the ska build process doesn't get as far as calling conda build.

**To Reproduce**
With the above compiler jinja, I end up with:
```
Using temporary directory /var/folders/63/ljv0yn5x01x6xf4cymvh814m0000gn/T/tmpb4r3xs7q for cloning
- Building package xija.
  - Cloning or updating source source xija.
xija failed, continue anyway (y/n)?
n
Traceback (most recent call last):
  File "ska_builder.py", line 166, in build_list_packages
    clone_repo(pkg_name, args, src_dir)
  File "ska_builder.py", line 70, in clone_repo
    meta = yaml.safe_load(jinja2.Template(meta_text).render())
  File "/Users/jean/conda_4.8.2/envs/builder/lib/python3.6/site-packages/jinja2/environment.py", line 1090, in render
    self.environment.handle_exception()
  File "/Users/jean/conda_4.8.2/envs/builder/lib/python3.6/site-packages/jinja2/environment.py", line 832, in handle_exception
    reraise(*rewrite_traceback_stack(source=source))
  File "/Users/jean/conda_4.8.2/envs/builder/lib/python3.6/site-packages/jinja2/_compat.py", line 28, in reraise
    raise value.with_traceback(tb)
  File "<template>", line 18, in top-level template code
jinja2.exceptions.UndefinedError: 'compiler' is undefined

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "ska_builder.py", line 207, in <module>
    main()
  File "ska_builder.py", line 203, in main
    build_list_packages(pkg_names, args, src_dir, build_dir)
  File "ska_builder.py", line 176, in build_list_packages
    raise ValueError(f"{pkg_name} failed")
```

**Additional context**

I think I can just call out the compiler pieces explicitly instead of using the jinja syntactic sugar (that's what I've been doing on the Perl skare3-static recipes anyway).  I suppose I could also try to do something *else* with the template so that by the time it gets to conda build it looks like {{ compiler('c') }} again.

taldcroft:
## Description

Allow there to be text like `{{ compiler('c') }}` in the `meta.yaml` file.

## Testing

- [N/A] Unit tests
- [x] Functional testing

Used this patch to build `sherpa` which contains:
```
requirements:
 build:
  - {{ compiler('cxx') }}
  - {{ compiler('c') }}
  - make
  - bison
  - flex
```

### Verification

Doing the build includes this output, which seems to indicate using local compilers:
```
+CLANG=$BUILD_PREFIX/bin/x86_64-apple-darwin13.4.0-clang
+CMAKE_PREFIX_PATH=:$PREFIX
+INDR=$BUILD_PREFIX/bin/x86_64-apple-darwin13.4.0-indr
+INSTALL_NAME_TOOL=$BUILD_PREFIX/bin/x86_64-apple-darwin13.4.0-install_name_tool
+LD=$BUILD_PREFIX/bin/x86_64-apple-darwin13.4.0-ld
...
+CLANGXX=$BUILD_PREFIX/bin/x86_64-apple-darwin13.4.0-clang++
+CXX=x86_64-apple-darwin13.4.0-clang++
```

Fixes #391

### DIFF
--- a/ska_builder.py
+++ b/ska_builder.py
@@ -67,7 +67,8 @@ def clone_repo(name, args, src_dir):
         return None
 
     # Stub out the jinja context variables and read meta.yaml
-    meta = yaml.safe_load(jinja2.Template(meta_text).render())
+    macro = '{% macro compiler(arg) %}{% endmacro %}\n'
+    meta = yaml.safe_load(jinja2.Template(macro + meta_text).render())
 
     # Upstream (home) URL is for the tags
     upstream_url = meta['about']['home']


jeanconn:
Does this mean we're also building sherpa?

jeanconn:
For testing, it would probably be good to just verify that you see the sherpa build use the anaconda compilers instead of system compilers (which would indicate that the right text made it to conda-build instead of being cut).

taldcroft:
@jeanconn - I put some outputs of the build into the description. Does that look good enough to you?

jeanconn:
I think so, though I don't really know if the activate script sets those even if the compiler is not used?  Later there is likely a line where `x86_64-apple-darwin13.4.0-clang++` actually does something (and it probably would have just been a naked `clang++` before or even a `g++` .

taldcroft:
And of course it's worth noting that this patch does not change anything about the way `conda build` behaves.  It is not actually changing the `meta.yaml` file, only allowing that meta to get parsed within `ska_builder.py`. So my testing was really testing `conda build`, which is fine, but it *should* be working as advertised.

PR #406, opened 2020-07-15
Add recipe for sherpa

taldcroft:
## Description

Add a recipe for building sherpa.  This is intended to be temporary and to be removed once an official conda package for Sherpa is available for Python 3.8 on linux and Mac.  The recipe was adapted with minimal modification from the recipe included in the sherpa source.

### MacOSX and Linux

Used `python ska_builder.py --python=3.8 sherpa` to successfully build `sherpa` 4.12.1.  This package was installed and passed `sherpa_test`.  I also ran `xija_gui_fit` using this and was able to successfully run a simple fit without any crash.

### DIFF
--- a/pkg_defs/sherpa/build.sh
+++ b/pkg_defs/sherpa/build.sh
@@ -1,0 +1,12 @@
+
+sed -i.orig "s|#install_dir=build|install_dir=$PREFIX|" setup.cfg
+git update-index --assume-unchanged setup.cfg
+
+export PYTHON_LDFLAGS=" "
+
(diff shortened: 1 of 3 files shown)


jeanconn:
As an aside, one reason I've been leaving around recipes we don't use is that they are then still github searchable.

taldcroft:
Maybe we can move them into some archive directory? Right now the default is to build every package that is in `pkg_defs`.  But more to the point I find it confusing to have recipes that we are not using.

PR #409, opened 2020-07-15
Move inactive recipes to inactive_pkg_defs for reference

taldcroft:
None of these recipes are being used currently but they are a useful reference and may someday be reactivated.


jeanconn:
The new ACIS packages are in ska3-flight but hadn't made it to the build list.

PR #421, opened 2020-07-22
Add recipe for prompt-toolkit

taldcroft:
This allows making a local build of 2.0.10 which fixes problems on Windows.

### DIFF
--- a/pkg_defs/prompt-toolkit/meta.yaml
+++ b/pkg_defs/prompt-toolkit/meta.yaml
@@ -1,0 +1,32 @@
+package:
+  name: prompt-toolkit
+  version:  {{ SKA_PKG_VERSION }}
+
+build:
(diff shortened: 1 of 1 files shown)


jeanconn:
Sorry you had to cross over to the dark side.

jeanconn:
What was the prompt-toolkit issue and is it likely to get fixed upstream?

taldcroft:
Oh dear, not a clue.

taldcroft:
Bad Tom for not providing a better trail of crumbs.

jeanconn:
No sweat, I was just noticing that we're handling sherpa and prompt-toolkit with this special handing... and while phasing out the sherpa special handling also wondered if there could come a time when this could go.  But I'd also guess there isn't rapid development on this package and it is unlikely to hold us back for a while (as other "core" packages advance).

PR #427, opened 2020-07-29
Tidy up some unused packages

taldcroft:
As far as I can see these packages are unused or overtaken.  I also decided I didn't like the `inactive_pkg_defs` name and changed to use `archived` in following suit with flight ops.


jeanconn:
The aca_egse package is still live, but it doesn't need to be frequently built from ska_builder.py and I don't care if it is in the archived packages or top level.

taldcroft:
Yeah, we may need to come back to that one (aca_egse).

PR #428, opened 2020-07-29
Shiny ska3 perl latest

taldcroft:
The analogy to ska3-flight-latest and ska3-core-latest.

### DIFF
--- a/pkg_defs/ska3-perl-latest/meta.yaml
+++ b/pkg_defs/ska3-perl-latest/meta.yaml
@@ -1,0 +1,25 @@
+package:
+  name: ska3-perl-latest
+  version: 2020.2
+
+build:
+  skip: True  # [win]
+
+requirements:
+  run:
+    - perl
+    - perl-app-cpanminus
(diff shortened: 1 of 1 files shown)


jeanconn:
I was figuring the libuuid thing would not be an issue requiring a pin in the metapackage because mostly conda-forge won't be in channels?  But I suppose ska3-perl-latest has a small enough user base (us) that it doesn't matter.

taldcroft:
It's needed to make the original process of assembling the packages (starting from scratch) reproducible / deterministic.

jeanconn:
OK. So does this replace the conda install package set in my build process or update a secondary process?

taldcroft:
Yes, this replaces the long command-line list and is the "latest" analog to the existing `ska3-perl` package. So now `conda create -n ska3-shiny -c ... ska3-core-latest ska3-flight-latest ska3-perl-latest`.

BTW did you ever file an issue with conda on the problem in the defaults perl package? Maybe they can just fix that!

PR #437, opened 2020-08-20
Changed some meta-data

javierggt:
- In all cases where home was a github repository, changed it to the corresponding https address.

I thought of also defining the doc_url and summary  fields (like I did with [skare3_tools](https://github.com/sot/skare3/blob/master/pkg_defs/skare3_tools/meta.yaml#L37)). Anyway the doc address will change after we overhaul the documentation. Still, I could do it...

### DIFF
--- a/pkg_defs/Chandra.Maneuver/meta.yaml
+++ b/pkg_defs/Chandra.Maneuver/meta.yaml
@@ -39,4 +39,4 @@ test:
 
 
 about:
-  home: git@github.com:sot/Chandra.Maneuver
+  home: https://github.com/sot/Chandra.Maneuver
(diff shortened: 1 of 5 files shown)


jeanconn:
Will this have the side-effect that if one tries to build manually from scratch one will need to type the password for each repo or some such?  Or no impact?

taldcroft:
On Mac and Windows at least, https seems to just work with no prompting.  I'm not sure exactly how though.  I had the converse problem building on kady because the ssh agent won't work through a plain terminal so I had to authenticate for each repo with ssh git address.

javierggt:
Good question. If it does have that side-effect, then the script should be changed, because the home should be a web address. One could use the dev_url for fetching.

This does not apply to all packages, only a few. Are these the only private ones?

jeanconn:
Yes, these are the private ones.

taldcroft:
I just confirmed on kady that if I provide my personal token once for https access, then it just works after that. So at some level that is better.

PR #441, opened 2020-10-06
bugfix: added numpy to cxotime meta.yml.

javierggt:
Fixes https://github.com/sot/cxotime/issues/19


taldcroft (approved):
Yup, this makes sense as it does need numpy to build now.

Issue #188, opened 2019-08-15
A production not-Perl metapackage?


jeanconn:
After installing acdc on GRETA ska3/flight, I wondered, perhaps there are more packages like this that do not really need to go in ska3-flight but belong in production?

We can think more about this and make a new metapackage if needed (or rename ska3-perl to ska3-production).

taldcroft:
:+1: on renaming ska3-perl to ska3-production.

About acdc, I can see the point there, though acdc does provide processing utilities that could be useful in generic ACA analysis.  In particular the different algorithms for combining quadrant readouts into cal images, e.g. one at a time or median filtering.  That's the sort of thing I would do standalone.

jeanconn:
So maybe we have some more interest in this, though figuring out which packages belong in a new ska3-production package is somewhat not trivial.  

I *think* this would mean on the Perl side, for example, we'd want ska3-perl (or some other name) to have the minimal set for starcheck to work and ska3-production to have the rest of the Perl stuff plus any isolated packages that are only useful on HEAD.

taldcroft:
I would favor an incremental approach that let's us move forward now without getting stuck by non-trivial decisions.  In other words get the low-hanging fruit with  an initial PR:
- Add aca_hi_bgd
- Move acdc from flight to production
- Any other completely uncontroversial packages.

jeanconn:
OK. I just thought even here you'd said there was value in keeping acdc in ska3-flight for the algorithms for combining quadrants.

taldcroft:
Yes, my former self had more insight. The point is just to make some kind of progress forward without overthinking it too much. It doesn't matter all that much. Making ska-production is fine, or just tossing another package into ska3-flight will only make it negligibly bigger and that's fine. Your choice.

javierggt:
Old issue, more than 2 years old. I think this discussion just died out.

In this issue you even agreed to rename ska3-perl to ska3-production. Two years later it has not happened and I do not see it happening yet 😄 I'm sure that can be considered again, but this discussion should be closed.

jeanconn:
Well, we did continue the discussion even up to October.  I suppose I don't know if open discussion best live in closed or open issues...

javierggt:
Feel free to reopen it if it is an ongoing discussion.

PR #443, opened 2020-10-14
Add perigee_health_plots and Ska.report_ranges

jeanconn:
Add perigee_health_plots and Ska.report_ranges

### DIFF
--- a/pkg_defs/Ska.report_ranges/meta.yaml
+++ b/pkg_defs/Ska.report_ranges/meta.yaml
@@ -32,5 +32,4 @@ test:
 
 
 about:
-  home: https://github.com/sot/aca_hi_bgd
-
+  home: https://github.com/sot/Ska.report_ranges
--- a/pkg_defs/Ska.report_ranges/meta.yaml
+++ b/pkg_defs/Ska.report_ranges/meta.yaml
@@ -1,0 +1,36 @@
+package:
+  name: ska.report_ranges
+  version:  {{ SKA_PKG_VERSION }}
+
+build:
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+
+source:
+  path: {{ SKA_TOP_SRC_DIR }}/Ska.report_ranges
+
+# the build and runtime requirements. Dependencies of these requirements
+# are included automatically.
+requirements:
+  # Packages required to build the package. python and numpy must be
+  # listed explicitly if they are required.
+  build:
+    - python
+    - setuptools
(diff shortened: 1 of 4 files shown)


taldcroft:
Are either of this intended to be part of ska3-flight?  Ska.report_ranges seems like a candidate, though I'm not sure quite where it gets used, and don't really mind either way. Seems like perigee_health_plots is another one for ska-production #188?

It seems like it would be a good idea to generally also update other relevant metapackage(s) at the time of making a recipe, assuming the intent of the recipe is to get the package installed reproducibly.

jeanconn:
I figured updating the metapackages was really a separate step compared to getting recipes to build conda packages to test first, as, for example, haven't figured out what versions go anywhere yet.  ska3-production looks like it would probably have fixed versions?

And yes, I figured Ska.report_ranges in ska3-flight and perigee_health_plots in ska3-production.

pkg_defs/Ska.report_ranges/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - chandra.time
+    - testr
+
+test:
+  imports:
+    - Ska.report_ranges
+
+
+about:
+  home: https://github.com/sot/aca_hi_bgd
```
jeanconn, line 35: Oops.

taldcroft:
It needs to be done as a documented and reproducible process (as in copy/paste these lines and it will work) in the shiny maintainers Wiki.

The hint about where a package comes from is in /proj/sot/ska/www/ASPECT_ICXC/ska3-conda/shiny/repodata_*.json. Each package in there has a key `is_ska` that indicates if that package was built by ska_builder or is OTS. 

However, that may be different from flight / core, so in fact your original idea of building the core env, getting packages, and then adding in flight-latest seems fine. It just needs to be supplemented by tools / instructions that do not require any thinking or opening an editor.

jeanconn:
I was asking in the context of this PR in which you seemed to be telling me to put Ska.report_ranges in ska3-flight.  It sounds like there will be no one-line edits to ska3-flight for a while and my adding Ska.report_ranges to ska3-flight-latest in this PR instead is probably the right thing.  *Then* some kind of process will be used to make ska3-flight.

taldcroft:
> my adding Ska.report_ranges to ska3-flight-latest in this PR instead is probably the right thing. Then some kind of process will be used to make ska3-flight.

Agree on both.

PR #486, opened 2020-10-19
Add aca_hi_bgd to build order

jeanconn:
Add aca_hi_bgd to build order

### DIFF
--- a/ska3_flight_build_order.txt
+++ b/ska3_flight_build_order.txt
@@ -44,6 +44,7 @@ find_attitude
 ska_sync
 ska.report_ranges
 perigee_health_plots
+aca_hi_bgd
 ska3-template
 ska3-pinned
 ska3-core


jeanconn:
I think pkgs need to be in there to actually get build from the workflow?  But maybe I am confused.  I have also forgotten how ska3_flight_arch_specific.txt is used.

javierggt:
> I think pkgs need to be in there to actually get built from the workflow?

Yes, in principle they need to be there. That is why the build failed when you tagged the release (https://github.com/sot/aca_hi_bgd/actions/runs/315973538). Now, after merging, one can go to that actions page and click "Re-run jobs" in the build workflow (I'll try that now).

javierggt:
ah, now I see you must have tried that. Now it failed because it did not find the runner.

taldcroft:
`ska3_flight_arch_specific.txt` was there to make it easier to build only arch-specific files on Windows or linux. But it may have been overtaken by just checking the meta file instead, so it could probably go away. From what I can see it isn't in the repo anymore.

jeanconn:
Do you mean used in the repo anymore?  I see it top level in master and I figured all these production packages probably need to be arch specific, so wasn't sure if I should be adding them.  But from a search I couldn't find any code uses of the file.

taldcroft:
About the build order file, I didn't realize there is still that dependency.  I thought somehow each package just gets built if there is a `build.yml` in the workflow directory. So now I'm confused.

jeanconn:
I was just reading the output from the build actions which was doing something like
```
python ska_builder.py --github-https --force --build-list ska3_flight_build_order.txt --tag bf9501d01fa76f2af1bc62d4e18ffba097cce5c6 aca_hi_bgd
```
but I don't know where that command lives in skare3_tools just now.

taldcroft:
Yes, I meant that it is not used by ska_builder.py any more, or even get referenced anywhere else in the repo.

jeanconn:
@javierggt regarding "did not find the runner", what does that mean?  Do I need to fix something?  I see the Update Conda Channel action has failed.

PR #490, opened 2020-10-22
WIP: Shiny ska3-core/flight meta.yml

javierggt:
The meta.yml in this PR should work on linux, osx and windows.

The files were produced as described here: https://github.com/sot/skare3/wiki/Shiny-Ska3-for-maintainers#assembling-package-meta.

### DIFF
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -101,8 +101,8 @@ package:
       - libgcc-ng ==9.1.0  # [linux]
       - libgfortran ==3.0.1  # [osx]
       - libgfortran-ng ==7.3.0  # [linux]
-      - libiconv ==1.16  # [osx]
       - libiconv ==1.15  # [win]
+      - libiconv ==1.16  # [osx]
       - liblief ==0.10.1
       - libllvm9 ==9.0.1
       - libpng ==1.6.37
@@ -171,6 +171,8 @@ package:
       - pkginfo ==1.5.0.1
       - pluggy ==0.13.1
       - prometheus_client ==0.8.0
+      - prompt-toolkit ==2.0.10  # [win]
+      - prompt-toolkit ==3.0.5  # [linux or osx]
       - prompt_toolkit ==2.0.10  # [win]
       - prompt_toolkit ==3.0.5  # [linux or osx]
       - psutil ==5.7.0
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -1,7 +1,7 @@
 ---
 package:
   name: ska3-core
-  version: shiny_2020.12
+  version: 2020.12+shiny
 
   requirements:
     run:
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -1,7 +1,7 @@
 ---
 package:
   name: ska3-core
-  version: 2020.11
+  version: shiny-2020.12
 
   requirements:
     run:
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -68,7 +68,7 @@ package:
       - imagesize ==1.2.0
       - importlib-metadata ==1.7.0
       - importlib_metadata ==1.7.0
-      - intel-openmp ==2020.1  # [linux or win]
+      - intel-openmp ==2019.4  # [osx]
       - intel-openmp ==2020.1  # [linux or win]
       - ipykernel ==5.3.3
       - ipyparallel ==6.3.0
@@ -102,7 +102,7 @@ package:
       - libgfortran ==3.0.1  # [osx]
       - libgfortran-ng ==7.3.0  # [linux]
       - libiconv ==1.16  # [osx]
-      - libiconv ==1.16  # [osx]
+      - libiconv ==1.15  # [win]
       - liblief ==0.10.1
       - libllvm9 ==9.0.1
       - libpng ==1.6.37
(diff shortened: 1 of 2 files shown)


pkg_defs/ska3-core/meta.yaml
```diff
@@ -1,214 +1,267 @@
...
+      - pcre ==8.44  # [linux or osx]
+      - pexpect ==4.8.0
+      - pickleshare ==0.7.5
+      - pillow ==7.2.0
+      - pip ==20.1.1
+      - pkginfo ==1.5.0.1
+      - pluggy ==0.13.1
+      - prometheus_client ==0.8.0
+      - prompt_toolkit ==3.0.5  # [linux or osx]
+      - prompt_toolkit ==3.0.5  # [linux or osx]
```
jeanconn, line 175: Not sure about the duplicate.

pkg_defs/ska3-core/meta.yaml
```diff
@@ -1,214 +1,267 @@
...
+      - ripgrep ==11.0.2  # [linux or osx]
+      - rope ==0.17.0
+      - ruamel_yaml ==0.15.87
+      - runipy ==0.1.5  # [win]
+      - scikit-learn ==0.23.1
+      - scipy ==1.5.0
+      - send2trash ==1.5.0
+      - setuptools ==49.2.0
+      - setuptools-scm ==4.1.2
+      - setuptools_scm ==4.1.2
```
jeanconn, line 222: Also doubled along with sip below

pkg_defs/ska3-flight/meta.yaml
```diff
@@ -1,62 +1,60 @@
...
+      - dpa_check ==3.0.0
+      - fep1_actel_check ==3.0.0
+      - fep1_mong_check ==3.0.0
+      - find_attitude ==3.4.0
+      - hopper ==4.5.0
+      - kadi ==5.3.0
+      - maude ==3.4.0
+      - mica ==4.22.0
+      - parse_cm ==3.5.2
+      - prompt-toolkit ==3.0.5  # [linux or osx]
```
jeanconn, line 34: Is this here because of something custom ska to make it?  Here and in core twice.

pkg_defs/ska3-core/meta.yaml
```diff
@@ -1,214 +1,267 @@
...
+      - sherpa ==4.12.1  # [linux or osx]
+      - sip ==4.19.8  # [osx]
+      - sip ==4.19.8  # [osx]
```
jeanconn, line 225: Also doubled but perhaps non-impacting.

pkg_defs/ska3-core/meta.yaml
```diff
@@ -1,214 +1,267 @@
...
+      - m2w64-gmp ==6.1.0  # [win]
+      - m2w64-libwinpthread-git ==5.0.0.4634.697f757  # [win]
+      - markupsafe ==1.1.1
+      - matplotlib ==3.2.2
+      - matplotlib-base ==3.2.2
+      - mccabe ==0.6.1
+      - menuinst ==1.4.16  # [win]
+      - mistune ==0.8.4
+      - mkl ==2020.1  # [linux or win]
+      - mkl ==2020.1  # [linux or win]
```
jeanconn, line 134: Double

pkg_defs/ska3-core/meta.yaml
jeanconn, line 175: Double.

pkg_defs/ska3-flight/meta.yaml
javierggt: it could be, like sherpa. I removed sherpa by hand.

javierggt:
The apparently repeated packages were packages that had different versions in windows and linux/osx. It was a bug in my script. Now the windows versions are correct.

The case of prompt-toolkit is the same as sherpa. We have a recipe for it, so I guess it should move to ska3-flight. There is also prompt_toolkit (with underscore) that is already in ska3-core.

pkg_defs/ska3-core/meta.yaml
javierggt: Good catch, for example, this should have been

```
sip ==4.19.13  # [linux or win]
sip ==4.19.8  # [osx]
```

Similar with the other repeated lines.

taldcroft:
The personal gist referenced in the Wiki should be put in to `skare3`, no?

javierggt:
> The personal gist referenced in the Wiki should be put in to `skare3`, no?

as you wish. I figured this was really a one-off.

pkg_defs/ska3-core/meta.yaml
```diff
@@ -1,214 +1,269 @@
+---
 package:
   name: ska3-core
-  version: 2020.5
+  version: shiny_2020.12
```
taldcroft, line 4: Maybe getting pedantic, but what about making this version compliant with standards (https://www.python.org/dev/peps/pep-0440/#local-version-identifiers).  For instance `2020.12+shiny`.

javierggt: done

taldcroft:
> I figured this was really a one-off.

The original goal was to update at least the key core packages to the latest around once a year. Not clear if that is possible or what we want, but presumably the script would be used each time.

pkg_defs/ska3-core/meta.yaml
javierggt: In the business of being pedantic... strictly speaking, this is not the intent of local versions:

> Local version identifiers are used to denote fully API (and, if applicable, ABI) compatible patched versions of upstream projects... The inclusion of the local version label makes it possible to differentiate upstream releases from potentially altered rebuilds by downstream integrators.

Still, I am updating the sanity checks in the CI build so any tag following PEP-0440 can be built, as long as there is some consistency between branch name, release tag name, and release type. This would allow building things like this, abusing the local version.

javierggt: and in the subject... just having a separate name (ska3-flight-shiny or similar) would save us all the abuse (and that is what I believe @jeanconn suggested at some point).

javierggt:
Closing this and continuing on #491

PR #508, opened 2020-10-30
ska3-perl for shiny release candidate

jeanconn:
Make a ska3-perl versioned metapackage for the shiny release candidate.  

Update combine_arch helper/utility script to with an option to subtract off the packages described by another platform-specific json file.  This new option was used to make the ska3-perl metapackage as the ska3-perl package should basically be the difference between the environment defined by (ska3-core ska3-flight ska3-perl-latest) and the one defined by (ska3-core ska3-flight).

### DIFF
--- a/pkg_defs/ska3-core/combine_arch.py
+++ b/pkg_defs/ska3-core/combine_arch.py
@@ -30,29 +30,29 @@ def main():
     ska_packages = [p['package'] for p in packages.get_package_list()
                     if p['package'] and p['package'] not in exceptions]
 
+    def get_env_packages(env_string):
+        """
+        Get platform label and dict of packages from json of supplied file.
+        """
+        try:
+            platform, filename = env_string.split('=')
+        except ValueError:
+            print(f' - skipped {env}')
+        print(f' + {platform}: {filename}')
+        with open(filename) as fh:
+            return platform, {p['name']: p for p in json.load(fh)}
+
     environments = {}
     print(f'Reading environments:')
     for env in args.env:
-        try:
-            platform, filename = env.split('=')
-        except ValueError:
-            print(f' - skipped {env}')
-            continue
-        print(f' + {platform}: {filename}')
-        with open(filename) as fh:
-            environments[platform] = {p['name']: p for p in json.load(fh)}
+        platform, env_packages = get_env_packages(env)
+        environments[platform] = env_packages
 
     subtract_environments = {}
     print(f'Reading environments to subtract:')
     for env in args.subtract_env:
-        try:
-            platform, filename = env.split('=')
-        except ValueError:
-            print(f' - skipped {env}')
-            continue
-        print(f' + {platform}: {filename}')
-        with open(filename) as fh:
-            subtract_environments[platform] = {p['name']: p for p in json.load(fh)}
+        platform, env_packages = get_env_packages(env)
+        subtract_environments[platform] = env_packages
 
     for platform in environments:
         remove_keys = []
--- a/pkg_defs/ska3-core/combine_arch.py
+++ b/pkg_defs/ska3-core/combine_arch.py
@@ -14,6 +14,7 @@ def parser():
     parser_.add_argument("--name", help="name of the package", default="ska3-core")
     parser_.add_argument("--version", help="new package version", default="")
     parser_.add_argument("--env", action="append", help="environment file", default=[])
+    parser_.add_argument("--subtract-env", action="append")
     parser_.add_argument("--out",
                          help="filename for output file with combined list of files"
                               " for use in metapackage ")
@@ -40,6 +41,29 @@ def main():
         print(f' + {platform}: {filename}')
         with open(filename) as fh:
             environments[platform] = {p['name']: p for p in json.load(fh)}
+
+    subtract_environments = {}
+    print(f'Reading environments to subtract:')
+    for env in args.subtract_env:
+        try:
+            platform, filename = env.split('=')
+        except ValueError:
+            print(f' - skipped {env}')
+            continue
+        print(f' + {platform}: {filename}')
+        with open(filename) as fh:
+            subtract_environments[platform] = {p['name']: p for p in json.load(fh)}
+
+    for platform in environments:
+        remove_keys = []
+        for package in environments[platform]:
+            if (platform in subtract_environments
+                    and package in subtract_environments[platform]
+                    and subtract_environments[platform][package]['version']
+                    == environments[platform][package]['version']):
+                remove_keys.append(package)
+        for package in remove_keys:
+            del environments[platform][package]
 
     package_names = sorted(set(sum([list(e.keys()) for e in environments.values()], [])))
     all_packages = []
--- a/pkg_defs/ska3-perl/meta.yaml
+++ b/pkg_defs/ska3-perl/meta.yaml
@@ -1,24 +1,42 @@
 package:
   name: ska3-perl
-  version: 2020.01.24
+  version: 2020.13+shiny
+
+build:
+  skip: True  # [win]
 
 requirements:
   run:
-    - cfitsio ==3.450
-    - xtime ==1.2.2
-    - pgplot ==5.2.2 1 # [linux]
+    - _sysroot_linux-64_curr_repodata_hack ==3  # [linux]
+    - cfitsio ==3.470  # [linux]
+    - gsl ==2.4  # [linux]
+    - kernel-headers_linux-64 ==3.10.0  # [linux]
+    - krb5 ==1.18.2  # [linux]
+    - libcurl ==7.71.1  # [linux]
+    - libssh2 ==1.9.0  # [linux]
+    - libx11-common-cos7-x86_64 ==1.6.7  # [linux]
+    - libx11-cos7-x86_64 ==1.6.7  # [linux]
+    - libx11-devel-cos7-x86_64 ==1.6.7  # [linux]
     - perl ==5.26.2
-    - perl-extutils-f77 ==1.20
     - perl-app-cpanminus ==1.7044
-    - perl-core-deps ==0.1
-    - perl-pgplot ==2.19 # [linux]
-    - perl-dbd-sybase ==1.15 pl526_1 # [linux]
-    - perl-chandra-time ==0.9.2
-    - perl-app-env-ascds ==0.4
-    - perl-cxc-sysarch ==1.0
-    - perl-ska-agasc ==3.5.0
+    - perl-app-env-ascds ==0.04  # [linux]
+    - perl-chandra-time ==0.9.2  # [linux]
+    - perl-core-deps ==0.3
+    - perl-cxc-sysarch ==1.00  # [linux]
+    - perl-dbd-sybase ==1.15  # [linux]
+    - perl-extended-deps ==0.1  # [linux]
+    - perl-ska-agasc ==3.5.0  # [linux]
     - perl-ska-classic ==0.6
-    - perl-ska-convert ==4.3
-    - perl-ska-web ==4.0
-    - watch_cron_logs ==4.2
-    - task_schedule ==4.2.1
+    - perl-ska-convert ==4.3  # [linux]
+    - perl-ska-web ==4.0  # [linux]
+    - perl-tk ==804.035  # [linux]
+    - pkg-config ==0.29.2
+    - sysroot_linux-64 ==2.17  # [linux]
+    - xorg-kbproto ==1.0.7  # [linux]
+    - xorg-libx11 ==1.6.9  # [linux]
+    - xorg-libxft ==2.3.2  # [linux]
+    - xorg-libxrender ==0.9.10  # [linux]
+    - xorg-renderproto ==0.11.1  # [linux]
+    - xorg-xproto ==7.0.31  # [linux]
+    - xtime ==1.2.2  # [linux]
+    - zip ==3.0
--- a/pkg_defs/ska3-perl/meta.yaml
+++ b/pkg_defs/ska3-perl/meta.yaml
@@ -7,6 +7,7 @@ build:
 
 requirements:
   run:
+    - ska3-core ==2020.13+shiny
     - _sysroot_linux-64_curr_repodata_hack ==3  # [linux]
     - cfitsio ==3.470  # [linux]
     - gsl ==2.4  # [linux]
--- a/pkg_defs/ska3-perl/meta.yaml
+++ b/pkg_defs/ska3-perl/meta.yaml
@@ -8,6 +8,8 @@ build:
 requirements:
   run:
     - ska3-core ==2020.13+shiny
+    - task_schedule ==4.2.1
+    - watch_cron_logs ==4.2
     - _sysroot_linux-64_curr_repodata_hack ==3  # [linux]
     - cfitsio ==3.470  # [linux]
     - gsl ==2.4  # [linux]


pkg_defs/ska3-core/combine_arch.py
```diff
@@ -41,6 +42,29 @@ def main():
...
+        with open(filename) as fh:
+            subtract_environments[platform] = {p['name']: p for p in json.load(fh)}
+
+    for platform in environments:
+        remove_keys = []
+        for package in environments[platform]:
+            if (platform in subtract_environments
+                    and package in subtract_environments[platform]
+                    and subtract_environments[platform][package]['version']
+                    == environments[platform][package]['version']):
```
jeanconn, line 63: This can obviously be cleaned up, but this is just the way my mind was working and seemed passable for a utility script.

javierggt: I think the idea of subtracting environments is nice, but does it produce a working package? This procedure produces the complement of ska3-core, but if the intersection of the packages is non-empty, then this will not be a complete package.

I think a requirement is that packages must work. In other words, installing a package should lead to an environment with all the intended tools. The only options then are:
- do this only with packages that do not intersect (do these two intersect?),
- duplicate the intersection of the packages (I am ok with this one),
- make one depend on the other.

taldcroft:
I don't have time to scrutinize the code now, but can you please provide a proper description for this PR of the rationale for the change and the intended use case.

jeanconn:
Little confused by what is a proper description given context? I just thought we wanted a ska3-perl metapackage with a set of package versions to go into shiny... but sure.

pkg_defs/ska3-core/combine_arch.py
jeanconn: I don't really understand the question... but part of the problem is likely that I also didn't understand your plan on or how I was supposed to use the current code to make a ska3-perl.  So, to quickly make a ska3-perl that included the package either brought in by ska3-perl-latest or its dependencies, I didn't:

conda create -n ska3-core-flight ska3-core ska3-flight --override-channels -c https://icxc.harvard.edu/aspect/ska3-conda/shiny

and from that make ska3-core-flight-linux.json and ska3-core-flight-osx.json

conda create -n ska3-core-flight-perl ska3-core ska3-flight ska3-perl-latest --override-channels -c https://icxc.harvard.edu/aspect/ska3-conda/shiny

and from that make ska3-core-flight-perl-linux.json and ska3-core-flight-perl-osx.json

And then used the tool like 

combine_arch.py --env linux=ska3-core-flight-perl-linux.json --env osx=ska3-core-flight-perl-osx.json 
  --subtract-env linux=ska3-core-flight-linux.json --subtract-env osx=ska3-core-flight-osx.json  --out meta.yaml

And then rename the meta.yaml package name because I didn't fiddle with the part writing out ska3-core.  So this kind of subtraction seems like it would be useful to make ska3-flight if you didn't already have the packages mechanism... and it *seems* to have put together a reasonable list of packages for the ska3-perl package in this PR.  But I haven't built and tested the package yet.

jeanconn:
This is a tiny PR against 2020.13+shiny.  I was originally going to push these two tiny commits directly to 2020.13+shiny branch, but figured that it would be a bit rude without letting Javier have a look.

In 2020.13+shiny there's a new combine_arch.py to make ska3-core.  That script has logic to take the package versions from two platforms, subtract ska packages, and make one list of everything left over.  We also have need for a platform-specific all-versioned package for ska3-perl made from the "extra" packages installed in ska3 by the ska3-perl-latest package, so some of the combine_arch code seemed useful.  It seemed to me that the easiest way to do that was to make an environment with ska3-core ska3-flight ska3-perl-latest, and subtract off all the packages already defined in an environment created just by ska3-core ska3-flight. so I added code to the modified combine_arch.py helper script to remove packages on a platform from a separate json.

For this PR, the question is really: should these changes end up in combine_arch.py or should I just toss it into a gist like Javier's at https://gist.github.com/javierggt/f64fd5300c1e946e9f3f52f483195e8c .  They probably don't belong in combine_arch if we leave it hanging around in the ska3-core directory.  It just seemed to me that it was becoming a bit more general

I've changed this to a WIP because I haven't tested both ska3-perl packages.

jeanconn:
I confirmed the ska3-perl packages build and look to work so I just removed the WIP.

javierggt:
@taldcroft, the bottomline is this:
- to set the ska3-perl versions in ska3-perl/meta.yaml.
- Jean is doing that by listing packages required by ska3-perl-latest which are not in a ska3-core-latest environment, and she made changes to the script to produce the list.

I suggested just listing packages installed by ska3-perl-latest (without removing packages in ska3-core) unless we make ska3-core a requirement of ska3-perl.

For example, if you create a fresh environment and install ska3-core-latest, you will have yaml, sqlite, numpy and others. If you install just ska3-perl you might not get those, because they come through ska3-core and ska3-core is not a requirement.

pkg_defs/ska3-core/combine_arch.py
javierggt: Sorry, I should just have done it. What I meant is to do:

on linux:
```
conda create -n ska3-perl --override-channels -c https://icxc.cfa.harvard.edu/aspect/ska3-conda/shiny ska3-perl-latest
conda activate ska3-perl
conda list --json > ska3-perl-linux.json
```

on OSX
```
conda create -n ska3-perl --override-channels -c https://icxc.cfa.harvard.edu/aspect/ska3-conda/shiny ska3-perl-latest
conda activate ska3-perl
conda list --json > ska3-perl-osx.json
```

then
```
./combine_arch.py --name ska3-perl --version 2020.13+shiny --env linux=ska3-perl-linux.json --env osx=ska3-perl-osx.json --out ska3-perl/meta.yaml
```

That produced this:
```
    - _libgcc_mutex ==0.1  # [linux]
    - _sysroot_linux-64_curr_repodata_hack ==3  # [linux]
    - blas ==1.0  # [linux]
    - bzip2 ==1.0.8  # [linux]
    - ca-certificates ==2020.6.24
    - certifi ==2020.6.20
    - cfitsio ==3.470  # [linux]
    - expat ==2.2.9
    - fontconfig ==2.13.0  # [linux]
    - freetype ==2.10.2  # [linux]
    - gsl ==2.4  # [linux]
    - icu ==58.2  # [linux]
    - intel-openmp ==2020.1  # [linux]
    - kernel-headers_linux-64 ==3.10.0  # [linux]
    - krb5 ==1.18.2  # [linux]
    - ld_impl_linux-64 ==2.33.1  # [linux]
    - libcurl ==7.71.1  # [linux]
    - libcxx ==10.0.0  # [osx]
    - libedit ==3.1.20191231
    - libffi ==3.3
    - libgcc-ng ==9.1.0  # [linux]
    - libgfortran-ng ==7.3.0  # [linux]
    - libiconv ==1.16  # [osx]
    - libpng ==1.6.37  # [linux]
    - libssh2 ==1.9.0  # [linux]
    - libstdcxx-ng ==9.1.0  # [linux]
    - libuuid ==1.0.3  # [linux]
    - libx11-common-cos7-x86_64 ==1.6.7  # [linux]
    - libx11-cos7-x86_64 ==1.6.7  # [linux]
    - libx11-devel-cos7-x86_64 ==1.6.7  # [linux]
    - libxcb ==1.14  # [linux]
    - libxml2 ==2.9.10  # [linux]
    - mkl ==2020.1  # [linux]
    - mkl-service ==2.3.0  # [linux]
    - mkl_fft ==1.1.0  # [linux]
    - mkl_random ==1.1.1  # [linux]
    - ncurses ==6.2
    - numpy ==1.18.5  # [linux]
    - numpy-base ==1.18.5  # [linux]
    - openssl ==1.1.1g
    - perl ==5.26.2
    - perl-app-cpanminus ==1.7044
    - perl-app-env-ascds ==0.04  # [linux]
    - perl-chandra-time ==0.9.2  # [linux]
    - perl-core-deps ==0.3
    - perl-cxc-sysarch ==1.00  # [linux]
    - perl-dbd-sybase ==1.15  # [linux]
    - perl-extended-deps ==0.1  # [linux]
    - perl-ska-agasc ==3.5.0  # [linux]
    - perl-ska-classic ==0.6
    - perl-ska-convert ==4.3  # [linux]
    - perl-ska-web ==4.0  # [linux]
    - perl-tk ==804.035  # [linux]
    - pip ==20.1.1
    - pkg-config ==0.29.2
    - python ==3.8.3
    - readline ==8.0
    - setuptools ==49.2.0
    - six ==1.15.0  # [linux]
    - sqlite ==3.32.3
    - sysroot_linux-64 ==2.17  # [linux]
    - tk ==8.6.10
    - wheel ==0.34.2
    - xorg-kbproto ==1.0.7  # [linux]
    - xorg-libx11 ==1.6.9  # [linux]
    - xorg-libxft ==2.3.2  # [linux]
    - xorg-libxrender ==0.9.10  # [linux]
    - xorg-renderproto ==0.11.1  # [linux]
    - xorg-xproto ==7.0.31  # [linux]
    - xtime ==1.2.2  # [linux]
    - xz ==5.2.5
    - yaml ==0.2.5
    - zip ==3.0
    - zlib ==1.2.11
```

pkg_defs/ska3-perl/meta.yaml
```diff
@@ -1,24 +1,42 @@
...
+    - perl-chandra-time ==0.9.2  # [linux]
+    - perl-core-deps ==0.3
+    - perl-cxc-sysarch ==1.00  # [linux]
+    - perl-dbd-sybase ==1.15  # [linux]
+    - perl-extended-deps ==0.1  # [linux]
+    - perl-ska-agasc ==3.5.0  # [linux]
     - perl-ska-classic ==0.6
-    - perl-ska-convert ==4.3
-    - perl-ska-web ==4.0
-    - watch_cron_logs ==4.2
```
jeanconn, line 23: I suppose I should add watch_cron_logs and task_schedule back in.

pkg_defs/ska3-core/combine_arch.py
jeanconn: Right.  I think a bunch of those packages are in ska3-core so I'm not thrilled with the overlap.

pkg_defs/ska3-perl/meta.yaml
jeanconn: They were removed by default by combine_arch used this way (because combine_arch is meant to make ska3-core and no ska packages belong in ska3-core).

pkg_defs/ska3-core/combine_arch.py
javierggt: I am sorry to keep saying this, but is ska3-core a requirement of ska3-perl? If ska3-perl is intended to always be used with ska3-core, then let's add the explicit requirement.

javierggt: if it is not, then those packages need to be there, and it means that ska3-perl should be generated whenever ska3-core is generated, to make sure there are no conflicts. I am ok with that, since we have a script to do it.

jeanconn: Indeed, in this PR I've ignored the fact that ska3-core is really a run requirement for ska3-perl.  

For our use cases (ska3-perl will only ever be installed with ska3-core and ska3-flight) I didn't really see the driver.  If we add ska3-core to the ska3-perl metapackage it just means we'd need to build new ska3-perl metapackages every time we updated ska3-core as you note.  I thought we could get away with fewer package builds.

pkg_defs/ska3-perl/meta.yaml
```diff
@@ -1,24 +1,42 @@
...
+    - pkg-config ==0.29.2
+    - sysroot_linux-64 ==2.17  # [linux]
+    - xorg-kbproto ==1.0.7  # [linux]
+    - xorg-libx11 ==1.6.9  # [linux]
+    - xorg-libxft ==2.3.2  # [linux]
+    - xorg-libxrender ==0.9.10  # [linux]
+    - xorg-renderproto ==0.11.1  # [linux]
+    - xorg-xproto ==7.0.31  # [linux]
+    - xtime ==1.2.2  # [linux]
+    - zip ==3.0
```
jeanconn, line 42: This could probably live in ska3-core-latest.

jeanconn:
If we aren't going to be making new ska3-core packages really often, I don't care, so I just added ska3-core to the draft ska3-perl metapackage.  I also don't know how much we need to refine process for updating the all-versioned metapackages.

So I don't know what we want to do with this PR.  I was anticipating a small amount of review and refinement from @javierggt  and then a merge into the development branch with the idea that @taldcroft will be reviewing the bigger PR on the 2020.13+shiny branch next week.

pkg_defs/ska3-core/combine_arch.py
```diff
@@ -41,6 +42,29 @@ def main():
         with open(filename) as fh:
             environments[platform] = {p['name']: p for p in json.load(fh)}
 
+    subtract_environments = {}
```
taldcroft, line 45: Every time you find yourself copy/pasting a section of code and changing a couple of variables, ask yourself if it can be refactored. I believe the answer here is yes.

pkg_defs/ska3-perl/meta.yaml
taldcroft: If this dependency got generated from the perl side, then it belongs here.

taldcroft (commented):
In general this looks OK and the meta.yaml looks reasonable.

My real comment which is more substantial is that we need the process of generating this file to be documented and reproducible.  In this case Jean needed to think about things and then make two more commits. We will go through this whole exercise again in maybe 2 years by which time all of this will be forgotten entirely (by me anyway). So no hand-edits, or if there are then they need to be carefully documented.

To that end, one path forward is:
- Move watch_cron_logs and task_schedule into ska3-flight. Honestly I really don't care that they get installed and won't run. There is *nobody* that will try to do this and get confused because they don't run. The user community for those packages is us three.
- Remove the dependence on ska3-core here. Indeed nothing will work without ska3-core. Is there any single user that will get tripped up by this? No.

What I am saying here is to emphasize process simplicity over the perfection of having all the dependencies correct for mix-n-match installation of different components. This latter feature of conda is not a requirement for Ska3 meta packages now and I haven't yet heard a compelling reason to support this (given the real *cost* of doing so).

With that I think of this meta package as being essentially a plug-in to Ska3. Of course it won't work if you don't have ska3-core installed, but do we care? My feeling is no.

If we take this approach then I think (??) that the current combine_arch.py will generate the correct file out of the box and we're done and having something simple to document/maintain and reproducible.

pkg_defs/ska3-core/combine_arch.py
jeanconn: Sure.  I think my line for looking twice is a little different from yours on a bad utility script but I can make this nicer.

jeanconn:
Awesome.  Thanks for the feedback Tom!

> My real comment which is more substantial is that we need the process of generating this file to be documented and reproducible.

Agree 100%.  I have notes but was planning to finalize them after reusing whatever we decided as a final script to remake the meta.yaml in a process that would end up at the bottom of the shiny twiki page with Javier's notes.


> Move watch_cron_logs and task_schedule into ska3-flight. Honestly I really don't care that they get installed and won't run. There is nobody that will try to do this and get confused because they don't run. The user community for those packages is us three.

Sure.  That's a path forward.  I think it is probably more trivial to update the script to make ska3-perl to include them correctly?  I But if you think it makes more sense for them to live in ska3-flight that's fine.  I know we'd gone back and forth on them.  I suppose I'd just need to start by cutting new versions of them after updating their conda recipes to remove all their dependencies.  Will test that out.

> With that I think of this meta package as being essentially a plug-in to Ska3. Of course it won't work if you don't have ska3-core installed, but do we care? My feeling is no.

Understood.  Though I did think that a good point that came up for me from the discussion with Javier is that if we update ska3-core, we probably need to test and cut a new version of ska3-perl anyway (because changes in the core packages can conflict or change what ends up in ska3-perl made from ska3-perl-latest).  Because of that, I don't feel strongly that keeping ska3-core *out* of ska3-perl saves us much (still need to do the process on an update), but also fine to strip it out again.

> If we take this approach then I think (??) that the current combine_arch.py will generate the correct file out of the box and we're done and having something simple to document/maintain and reproducible.

Well, technically the script now puts out something called ska3-core at the top (and lives in the ska3-core directory), so it either needs a few additional tiny tweaks or documented steps... but will get the ducks in a row.

javierggt:
I guess we understand "simplicity" differently.

What I suggested is entirely reproducible in an automated way, and required:
- no change in the current code,
- updating ska3-perl whenever ska3-core is updated, which might be needed anyway.

Leaving out ska3-core from the dependencies appears to decouple the two packages so you can update one and not the other, but in reality there will be implicit dependencies, so ska3-perl might need updating whenever ska3-core is updated anyway.

I don't know why Jean had to manually add packages in the end. Were they not already in the inout environments?

jeanconn:
> I don't know why Jean had to manually add packages in the end. Were they not already in the input environments?

I was pretty sure they got excluded because they were ska packages and the current ska3-core/combine_arch is set to exclude ska packages via the skare3_tools packages mechanism.  So the simplest thing would be to modify this combine_arch draft script to not exclude packages via that mechanism when a subtracting off an environment supplied by the subtract-env option (it shouldn't do both).  

I'm also still not sure if the best plan is to make combine_arch perfect or just exactly capture what we did (like how to make ska3-flight) in a separate script / gist.

taldcroft:
About the ska3-core, I got misled by the commit where `ska3-core=2020.13+shiny` was added by hand. Agreed that this is fine if it just comes out of the process from the existing code. Sorry for causing confusion.

jeanconn:
I don't think we are *all* the way there yet with this script and the
interaction between ska-perl and ska3-core?  But I'm on-board with the
though that it doesn't matter if ska3-core is in the ska3-perl metapackage
as long as we have a reproducible process with very few or no manual edits.

javierggt:
> I was pretty sure they got excluded because they were ska packages and the current ska3-core/combine_arch is set to exclude ska packages via the skare3_tools packages mechanism. So the simplest thing would be to modify this combine_arch draft script to not exclude packages via that mechanism when a subtracting off an environment supplied by the subtract-env option (it shouldn't do both).

Makes sense, and I would agree with that.

Strictly speaking, we could do away with the skare3_tools.packages option and create ska3-core by installing ska3-flight-latest and subtracting ska3-core-latest. This means the procedure would be the same for both ska3-core and ska3-perl, and all the arbitrariness is absorbed in the creation of ska3-*-latest. Then, if you add ska3-core-latest to the dependencies of ska3-perl-latest, you get a ska3-core dependency in the end.

jeanconn:
I haven't checked, will your version RC rewriting stuff work to update the ska3-core version as a dependency?

jeanconn:
I suppose it is set to do that in ska3-flight already but wasn't sure where that was configured so it would work in a new metapackage like ska3-perl.

javierggt:
> I haven't checked, will your version RC rewriting stuff work to update the ska3-core version as a dependency?

no, but it would be an easy addition. I would add something that replaces all appearances of ska3-*latest by ska3-* of the same version being generated. This would remove some arbitrary steps:
- one would not have to remove ska3-core
- we would not have the special cases of packages in ska3_tools.packages not in ska3-flight (sherpa et al.)

jeanconn:
I'm still not sure that ska3-core in ska3-perl wins us much.  We'd add ska3-core-latest to ska3-perl-latest for testing, but would still need a manual step (or exception/special case in the code) to add ska3-core-latest (or ska3-core at a version to be magically updated) to ska3-perl.

javierggt:
actually... ska3-core would appear in both the ska3-flight and ska3-core environments, so by subtracting them it would not be in the final ska3-flight env. But still, one can add exceptions to ska3-packages (to not subtract them, for example).

javierggt:
My suggestion is to merge this, and I will add the option to handle `ska3-*`.

We all agree the process should be straight-forward and reproducible, and it is going to be faster to implement it than to discuss it. Then we can discuss based on the implementation.

javierggt (approved):

jeanconn:
Sure.  I'll just clean up that one part that Tom called out as DRY .  Then we're left with 

- [ ] possibly removing the skare3_tools packages mechanism altogether, setting it by option, or setting it to work when subtract-env not supplied
- [ ] adding an exception to the subtracting for ska3- packages
- [ ] adding some way to label the created metapackage
- [ ] maybe moving this script out of ska3-core

pkg_defs/ska3-core/combine_arch.py
jeanconn: Refactored.

taldcroft:
I just did that DRY one.

taldcroft:
See  34bdad0.

taldcroft:
Sorry I should have just left this alone and let you do it.

taldcroft:
Though I will argue my refactor is cleaner.

jeanconn:
Haha.  I was just going to tell you yours was one step cleaner, but you already took credit...

taldcroft:
So you should *practice* DRY on every script, even ones that you consider to be low-grade utility scripts.

PR #514, opened 2020-11-05
add sphinx-argparse

### LINKED ISSUE #513: Add package sphinx-argparse to ska3-core
sphinx-argparse is a good package to have. It generates the documentation shown [here](https://sot.github.io/skare3_tools/scripts.html)
These are the packages that would be added for this:
- sphinx-argparse-0.2
- commonmark-0.9.1
- future-0.18.2

javierggt:
This pulls in the following:
- sphinx-argparse-0.2
- commonmark-0.9.1
- future-0.18.2 

Fixes #513

### DIFF
--- a/pkg_defs/ska3-core-latest/meta.yaml
+++ b/pkg_defs/ska3-core-latest/meta.yaml
@@ -55,6 +55,7 @@ requirements:
    - setuptools
    - setuptools_scm
    - sherpa # [not win]
+   - sphinx-argparse
    - six
    - sphinx
    - sqlalchemy # [win]


jeanconn:
Should something like this be also marked as not for windows?  or does it / should it just work and not be an issue (I haven't been booting up my VM to test all windows things and don't know how much we should).

javierggt:
> Should something like this be also marked as not for windows? or does it / should it just work and not be an issue (I haven't been booting up my VM to test all windows things and don't know how much we should).

I thought about it only *after* I copied all the conda files into the channels. Strictly speaking, we only need sphinx-argparse in linux, where docs are built. On the other hand, right now I made skare3_tools depend on it, so it would be needed wherever we want skare3_tools. I could remove the dependency though.

taldcroft:
I build docs on Mac at least for preview / correctness. Not sure if this would be needed there, but we should include at least Mac / linux.

jeanconn:
Right, I figured linux/mac made good sense but not sure if letting core packages creep more in windows just gave us more maintenance issues later.  But this one (3) seem reasonable for windows too.

javierggt:
well, I included all of them.

PR #530, opened 2020-11-12
change chandra_aca build requirement to include astropy

javierggt:


taldcroft:
I realize now this could be (maybe *should* be) done with requests instead of astropy.  I just had that code from the original version when it was being done in the user code not setup.py. Probably doesn't matter, but requests is a much lighter dependency.

taldcroft:
But certainly we can run with this for now.

javierggt:
well, the chandra_aca is failing anywya, and I have never seen that... I'll have lunch and try again in about half hour.

javierggt:
so it might still be a good idea to change the dependency.

PR #598, opened 2021-02-03
Add plotly and tqdm to ska3-core-latest

taldcroft:
These two packages should be capture as core dependencies.  `tqdm` is currently coming in as a derived dependency, so this makes it a primary one given direct use in Ska packages now.


jeanconn (approved):
LGTM

PR #600, opened 2021-02-04
Add jobwatch

jeanconn:
Add jobwatch


javierggt:
fine by me.

javierggt:
I can remove the one I added in the release branch, since this might be better. Also, mine does not work.

The package is arch-dependent?

jeanconn:
It has a couple of entry-point scripts so I think that makes it arch-dependent.

javierggt:
ok, let's merge 😄

PR #615, opened 2021-03-09
Add env vars for HEAD Sybase mica test

jeanconn:
Add env vars for HEAD Sybase mica test

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -164,3 +164,6 @@ jobs:
         env:
           CONDA_PASSWORD: ${{ secrets.CONDA_PASSWORD }}
           SKA: /proj/sot/ska
+          SYBASE_OCS: OCS-16_0
+          SYBASE: /soft/SYBASE16.0
+          LD_LIBRARY_PATH: /soft/SYBASE16.0/OCS-16_0/lib


jeanconn:
@javierggt if you don't want to source into a ska environment with ska_envs.sh, maybe we could just set these vars to let the mica report test run in your HEAD test environment?

javierggt:
I think this `LD_LIBRARY_PATH` gets ignored in the environment.

I tried this once and it did not work, possibly because I still had no access. We can try again, but when I try, I need to remind myself how is LD_LIBRARY_PATH set (and document it somewhere).

jeanconn:
IIRC setting the variable worked for me in a local test, but I didn't have a good way to test it with the workflow.

jeanconn:
I still thought this would probably work but am now confused about what permissions the runner has.

javierggt (approved):
I created a [test workflow](https://github.com/sot/skare3/runs/2796698431?check_suite_focus=true) that defines these environmental variables, and test_get_rms_data passes. It seems to work, so I say we merge this.

Just keep in mind that the environment of the runner was one of the things I struggled a bit when setting things up, and I have not repeated it. I should do that soon, so it is documented precisely.

PR #616, opened 2021-04-01
Add fot_matlab module

jeanconn:
Add fot_matlab Python package as a fot-matlab conda package.

### DIFF
--- a/pkg_defs/fot-matlab/meta.yaml
+++ b/pkg_defs/fot-matlab/meta.yaml
@@ -1,12 +1,12 @@
 package:
-  name: fot_matlab
+  name: fot-matlab
   version:  {{ SKA_PKG_VERSION }}
 
 build:
   script: python setup.py install --single-version-externally-managed --record=record.txt
 
 source:
-  path: {{ SKA_TOP_SRC_DIR }}/fot_matlab
+  path: {{ SKA_TOP_SRC_DIR }}/fot-matlab
 
 # the build and runtime requirements. Dependencies of these requirements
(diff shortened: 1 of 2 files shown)


jeanconn:
This isn't working for me yet with ska_builder, so still a wip.

jeanconn:
I just don't know if we want this a conda package fot-matlab or fot_matlab . Makes no nevermind to me.

javierggt:
fot-matlab

jeanconn:
OK I renamed the pieces that were fot_matlab to fot-matlab .  Not knowing the storied history of modules with dashes instead of underscores, I thought it was just a typo to have the dash in the name before, but if it is liked that is fine.

jeanconn:
I'll wait to merge until this has had a review.

javierggt:
Does this package define any scripts? It does not seem like it. Does it have any platform dependent component?

If the answer to both of these is no, then this should be a noarch package:
```
build:
  noarch: python
```

jeanconn:
Sure.  I suppose I was half thinking that with your automated build system noarch packages didn't buy us anything except a problem later when a script or task_schedule or some such gets added and then we can't remember why it doesn't work!

javierggt (approved):

javierggt:
well, I guess you are right about that last point.

javierggt:
it's just that when I added the workflow, I was half thinking it would be noarch :-) the two need to agree.

jeanconn:
Sounds good.

PR #674, opened 2021-07-08
Xija linux build

javierggt:
This PR changes the recipe for the xija package. It fixes the current build issue by building the package in python 3.8.10 and also fixing the versions of other build dependencies.

I verified that this:
- [x] builds in the upcoming version of the docker builder
- [x] builds in the current version of the docker builder
- [x] installs in our current installation without adding/updating packages to ska3-core
- [x] passes unit tests on HEAD

To build from this branch using the current docker builder image:
```
docker run --rm  --name builder -v $GITHUB_WORKSPACE:/github/workspace -w /github/workspace \
>     -e CONDA_PASSWORD -e GIT_USERNAME -e GIT_PASSWORD \
>     docker.pkg.github.com/sot/skare3/centos7-builder:latest \
>     sot/xija --skare3-branch linux-build --force --tag 4.23.0
```

Fixes #674

### DIFF
--- a/pkg_defs/Chandra.Time/meta.yaml
+++ b/pkg_defs/Chandra.Time/meta.yaml
@@ -14,9 +14,10 @@ requirements:
   # Packages required to build the package. python and numpy must be
   # listed explicitly if they are required.
   build:
-    - gxx_linux-64 # [linux]
+    - gcc_linux-64==8.4.0 # [linux]
+    - libgcc-ng==9.1.0  # [linux]
     - clangxx_osx-64 # [osx]
-    - python
+    - python==3.8.10
     - setuptools
     - setuptools_scm
     - setuptools_scm_git_archive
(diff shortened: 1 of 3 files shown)


jeanconn:
This seems reasonable to me.  We may not need the build number update (seems unlikely we will re-release a previous xija version with the new build).  Also, does this fix work for the other compiled packages?

javierggt:
> We may not need the build number update

the thing is... the package was already in the conda channel with build 0, so I needed it for this build. Of course, now that I manually uploaded the package, the number is not needed anymore.

> does this fix work for the other compiled packages?

I was thinking the same yesterday. It should work for all. The question is which other packages are there:
- cxotime
- ?

taldcroft (approved):
If this solves the problem (as evident from testing) then fine by me.

jeanconn:
For other packages, I'd figure Ska.Numpy.

jeanconn:
And with regard to build number, sure I can see that you needed it to make a test package to reuse in the test channel, but we probably don't need to increment the one in skare3 master.

PR #774, opened 2022-01-27
2022.2

javierggt:
# ska3-flight 2022.2

This PR includes:
- 

## Testing:

- [Automated tests](https://icxc.cfa.harvard.edu/aspect/skare3/testr/releases/2022.2/).

The latest release candidates will be installed in `/proj/sot/ska3/test` on HEAD, and all release candidates will be available for testing from the usual channels:
```
conda create -n ska3-flight-2022.2rc# --override-channels \
  -c https://icxc.cfa.harvard.edu/aspect/ska3-conda/flight \
  -c https://icxc.cfa.harvard.edu/aspect/ska3-conda/test \
  ska3-flight==2022.2rc#
```

## Review

All operations critical or impacting PR's are independently and carefully reviewed. For other PR's the level of detail for review is calibrated to operations criticality. Some PR's that are confined to aspect-team-specific processing may have little to no independent review.

## Deployment

ska3-flight 2022.2 will be promoted to flight conda channel and installed on HEAD and GRETA Linux after this week's loads are approved.

# Code changes

# Related Issues

### DIFF
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -1,289 +1,401 @@
 ---
 package:
   name: ska3-core
-  version: 2021.8  # in a typical release, change the ska3-core version in ska3-flight requirements.
+  version: 2022.2
 
 requirements:
   run:
     - _libgcc_mutex ==0.1  # [linux]
+    - _openmp_mutex ==4.5  # [linux]
     - alabaster ==0.7.12
-    - appnope ==0.1.0  # [osx]
-    - asgiref ==3.2.10
-    - astroid ==2.4.2
-    - astropy ==4.2.1
-    - astropy-healpix ==0.5
+    - alsa-lib ==1.2.3  # [linux]
+    - anyio ==2.2.0
+    - appdirs ==1.4.4
+    - appnope ==0.1.2  # [osx]
+    - argon2-cffi ==20.1.0
+    - asgiref ==3.4.1
+    - astroid ==2.9.0
+    - astropy ==5.0
+    - astropy-healpix ==0.6
     - astropy-sphinx-theme ==1.1
+    - astroquery ==0.4.5
+    - async_generator ==1.10
     - atomicwrites ==1.4.0  # [win]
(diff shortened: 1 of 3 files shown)


taldcroft:
The instructions for creating the environment don't work, where the first error I see is that `proseco-5.5.0` is not available.
After some playing around I found this would resolve everything, but it probably isn't right.
```
mamba create -n ska3-flight-2022.2rc3 --override-channels \
 -c https://icxc.cfa.harvard.edu/aspect/ska3-conda/next  \
-c https://icxc.cfa.harvard.edu/aspect/ska3-conda/test \
-c https://icxc.cfa.harvard.edu/aspect/ska3-conda/masters \ 
  ska3-flight==2022.2rc3 ska3-core=2022.2rc3
```

taldcroft:
Confirmed it works now thanks. In the boilerplate create command (`conda create -n ska3-flight-2022.2rc# ...`) in the description can you replace conda with mamba (in general)? That will make would-be testers happier.

taldcroft:
And sorry, not sure where to comment on the template, but "ska3-flight 2022.2 will be promoted to flight conda channel and installed on HEAD and GRETA Linux after this week's loads are approved." should be more like "ska3-flight 2022.2 will be promoted to flight conda channel and installed on HEAD and GRETA Linux at a time outside of the load review cycle."

Since FSDS is not tied to load review, all we do now is not promote between when loads are released and when they are approved.

javierggt:
> in the description can you replace conda with mamba (in general)?

That will work only if they have mamba installed already, which usually means you have a prior version of ska3-flight. I can change it, because it would work for us three. Just something to keep in mind.

javierggt:
> And sorry, not sure where to comment on the template

an issue. Also the template for release issues requires update.

taldcroft:
> That will work only if they have mamba installed already

Ah right. Never mind.

jeanconn:
Though maybe we should go ahead and make a "clean install" a two step procedure that installs mamba first an then mamba installs the rest.

PR #793, opened 2022-03-10
Remove deprecated acis packages from latest

jeanconn:
Remove deprecated acis packages from latest.

This is not tested as such, as it doesn't seem like this drives even a test release.


javierggt:
This is already in 2022.2.

jeanconn:
Interesting.  I figured if this were in the flight release then when I installed ska3-flight-latest today it wouldn't have grabbed a metapackage that required those old acis packages.  What kicks off an updated package build for this one?

jeanconn:
I installed to a clean environment with no versions, from flight and masters channels.  I'll try again.

PR #794, opened 2022-03-10
Add patch instruction files

javierggt:
This PR adds one `patch_instructions.json` file for each platform, which reflects what is currently on `ska3-conda/flight`.

We need to deal with package patches, there is no way around it. We have a few options:
- version control the final `patch_instructions.json` files (what this PR is about)
- have a python script to generate `patch_instructions.json` files ([what Anaconda does](https://github.com/AnacondaRecipes/repodata-hotfixes/))
- Modify the package themselves ([what we first tried](https://github.com/ContinuumIO/anaconda-issues/issues/11920))

The approach I took in producing the 2022.2 release was to run a workflow to:
- assemble a valid environment (using the `install_from_scratch.py` script in each metapackage),
- store a file with package metadata, including URL, as an artifact,
- fetche patch instructions from all upstream repositories,
- combine patch instructions for downloaded packages in a collection of json files (one per platform),
- store the patch instructions as an artifact,
- combine the package mata-data files (one per platform per package) into one `meta.yaml` per package, with the proper platform-dependent keywords.
- store the `meta.yaml` files as artifacts.

Because this is the first time I followed this process, the resulting patch instructions are what is included in this PR.

In future updates, if we need to update the patch instructions, it is necessary to add them to the repository. This can be done using the `-p` or `--patch-generator` option of conda index:

```
tar cf bundle.tar.bz2 -C ./temp .
conda index -p bundle.tar.bz2 www/ASPECT/ska3-conda/test
```

This should make the following two scripts obsolete, and I would remove both:
- [upload_packages.py](https://github.com/sot/skare3/blob/master/upload_packages.py)
- [fix_repodata.py](https://github.com/sot/skare3/blob/master/fix_repodata.py), already deprecated after Tom made `upload_packages.py`

### DIFF
--- a/patches/linux-64/patch_instructions.json
+++ b/patches/linux-64/patch_instructions.json
@@ -1,0 +1,141 @@
+{
+  "packages": {
+    "_openmp_mutex-4.5-1_llvm.tar.bz2": {
+      "license_family": "BSD"
+    },
+    "blosc-1.21.0-h9c3ff4c_0.tar.bz2": {
+      "license_family": "BSD"
+    },
+    "conda-build-3.21.7-py38h06a4308_0.tar.bz2": {
+      "depends": [
+        "beautifulsoup4",
+        "chardet",
+        "conda >=4.5",
+        "conda-package-handling >=1.3",
+        "filelock",
+        "glob2 >=0.6",
+        "jinja2 <3.0.0a0",
+        "patchelf",
+        "pkginfo",
+        "psutil",
+        "py-lief",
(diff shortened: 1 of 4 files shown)


jeanconn:
I think this is fine and doesn't need line-by-line review (proof of the pudding is in the eating .. re 2022.2).  I do wonder if we would be better served by moving the old scripts into an archived directory than deleting them.

jeanconn (approved):

Issue #73, opened 2018-08-13
Prune flt_envs


jeanconn:
For skare3, I think we can get rid of PER5LIB and LD_LIBRARY_PATH in flt_envs, but this will need to be tested on the Perl side some more.

javierggt:
What kind of tests are needed for this?

PR #855, opened 2022-05-23
Put an arc5gl in ska3-template and set for RH8

jeanconn:
Put an arc5gl in ska3-template and set for RH8

### DIFF
--- a/pkg_defs/ska3-template/bin/arc5gl
+++ b/pkg_defs/ska3-template/bin/arc5gl
@@ -1,0 +1,1199 @@
+#!/usr/bin/env %{PREFIX}%/bin/perlwrap
+
+# --8<--8<--8<--8<--
+#
+# Copyright (C) 2006 Smithsonian Astrophysical Observatory
+#
+# This file is part of src
+#
+# src is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# src is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the 
+#       Free Software Foundation, Inc. 
+#       51 Franklin Street, Fifth Floor
+#       Boston, MA  02110-1301, USA
+#
+# -->8-->8-->8-->8--
+
+
+use strict;
+use warnings;
+use Term::ReadLine;
+use Getopt::Long;
+use English;
+
+use CXC::Archive;
+use Chandra::Time;
+
+our $VERSION = '1.1.2_ska';
+
+############################################################################
+# This is a fake package to deal with test mode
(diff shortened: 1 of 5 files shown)


jeanconn:
I think we've hesitated in the past to remove $SKA/lib from LD_LIBRARY_PATH because we didn't define a test plan for the perl code https://github.com/sot/skare3/issues/73 or anything else that might by dynamically linked and need those $SKA/lib libraries.  

So I suppose this would need more comprehensive testing for Replan Central, arc5gl, and task_schedule to start.

I don't care about needing to support the other perl code (aca_egse splat) if we can just run them from an old environment and certify aca_view.

javierggt:
Why is `perlwrap` needed here?

jeanconn:
I duplicated the functionality that was in the previous arc5gl that called a perl that was always configured to have the skare3 environment variables from the #! without requiring that the user be "in" a skare3 environment, and I couldn't figure out how to do that without a wrapper.  If we don't need to functionality we can remove it, but I also figured it didn't cost much.

javierggt:
I figured that is what the `skare` script is for, but I don't know where `perlwrap` is used.

jeanconn:
It looks like, technically, RH8 supports the use of the "-S" flag to env, so this would work:
```
#!/usr/bin/env -S /proj/sot/ska3/flight/bin/skare perl
```
But that doesn't work with /usr/bin/env in CentOS7.

taldcroft:
> Can you think of a need to have arc5gl run directly without having HEAD ska environment variables sourced?

I personally would never think to do that. I'd type `ska3` like I do a dozen+ times a day and then type `arc5gl`. Is that your question?

- telem_archive: we really need to just track down if that is doing anything any more.
- fid_drift_mon: time to port to ska3.

jeanconn:
Maybe that is my question... What are you doing now to run arc5gl because I figure you aren't sourcing into ska2?  Or just doesn't come up?

taldcroft:
I would use my `ska` alias to source in to Ska2. But in practice I am using arc5gl manually only very rarely these days.

taldcroft:
Some quick research:
- telem_archive: we really need to just track down if that is doing anything any more.
  - This creates an `observations` table that does get used in a few places. It should be straightforward to use kadi commands v2 `get_observations` instead, but it will take some effort.
- fid_drift_mon: time to port to ska3.
  - The core data collection here is an old perl script. It would be best to port to Python but that is probably a day or two of work end to end.

javierggt:
Reading this thread... it seems like if this solves the issue, then we should merge, right? (and the thing about telem_archive and fid_drift_mon is a digression).

jeanconn:
Yes, the handling of the other perl stuff is a digression.  Since neither of you seem to want the perl wrapper in arc5gl I've removed it, though that does mean I'll need to redo my testing thus far.

taldcroft:
To be clear, I have no strong opinions about implementation, except:
- It works on CentOS-7 and RH8
- We get this done in finite time. :smile:

jeanconn:
I think my testing of this (which was basically against the draft environment) is sufficient.

jeanconn:
OK.  Merged.

jeanconn:
I'm not sure how things are set to kick off a new build of ska3-template, otherwise it can be built manually.

taldcroft:
Well this is exciting, long time coming!!

PR #864, opened 2022-05-31
add aimpoint_mon package

javierggt:
As the title says...


jeanconn (approved):
LGTM as an arch package.

PR #873, opened 2022-06-15
add kadi-apps

javierggt:
This PR adds a new package: kadi-apps. This packages is not intended to be part of ska3-flight/core.


javierggt:
Yes, I actually built the package locally, then copied it to a local repo on web-kadi-test, installed it there and then used it in the test env.

PR #921, opened 2022-09-16
add kalman watch

javierggt:


jeanconn (approved):
This looks fine to me, though I figure kalman_watch isn't going to run with the task_schedule in the package data unless we also make a release that includes https://github.com/sot/task_schedule/commit/bcc564f32a7f197f711b9f4ac932ab8809d8e00b .  Right?

javierggt:
That's a good point, and task_schedule goes in ska3-template, so I guess we need that.

javierggt:
sorry, ska3-perl

PR #977, opened 2022-12-20
add ska_helpers to build recipes

javierggt:
These changes will be needed in all current namespace packages if they are modified like in https://github.com/sot/Ska.Sun/pull/22

I noticed that many packages include this in their `setup.py`:
```python
try:
    from testr.setup_helper import cmdclass
except ImportError:
    cmdclass = {}
```

Up to now, this is never used, because testr is not a build dependency. This PR does not change that.


jeanconn (approved):
Looks good to me.  I think was already technically a run dependency (so it is somewhat a defect to not be in that list for each) but we don't manage the conda dependencies that closely.

PR #978, opened 2022-12-21
Update Ska.Sun to ska_sun

jeanconn:
Update Ska.Sun to ska_sun.

### DIFF
--- a/pkg_defs/ska_sun/meta.yaml
+++ b/pkg_defs/ska_sun/meta.yaml
@@ -1,6 +1,6 @@
 
 package:
-  name: ska.sun  # lower case name of package, may contain '-' but no spaces
+  name: ska_sun  # lower case name of package, may contain '-' but no spaces
   version: {{ SKA_PKG_VERSION }}
 
 
@@ -9,7 +9,7 @@ build:
   noarch: python
 
 source:
-  path: {{ SKA_TOP_SRC_DIR }}/Ska.Sun
+  path: {{ SKA_TOP_SRC_DIR }}/ska_sun
 
 
 
@@ -20,16 +20,20 @@ requirements:
     - setuptools_scm
     - setuptools_scm_git_archive
     - ska_helpers
+    - testr
   run:
     - python
     - ska.quatutil
     - chandra.time
     - quaternion
     - numpy
+    - ska_helpers
+    - testr
 
 test:
   imports:
     - Ska.Sun
+    - ska_sun
 
 about:
-  home: https://github.com/sot/Ska.Sun
+  home: https://github.com/sot/ska_sun
--- a/pkg_defs/Chandra.Maneuver/meta.yaml
+++ b/pkg_defs/Chandra.Maneuver/meta.yaml
@@ -28,7 +28,7 @@ requirements:
     - six
     - quaternion
     - chandra.time
-    - ska.sun
+    - ska_sun
     - ska_helpers
 test:
   requires:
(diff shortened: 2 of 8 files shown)


jeanconn:
@javierggt @taldcroft Do we want to update the conda package to be ska_sun or leave it as ska.sun?  I forget which package we decided to end-to-end test but could try this one...

pkg_defs/ska_sun/meta.yaml
```diff
@@ -20,16 +20,20 @@ requirements:
...
     - ska_helpers
+    - testr
   run:
     - python
     - ska.quatutil
     - chandra.time
     - quaternion
     - numpy
+    - ska_helpers
+    - testr
```
taldcroft, line 31: `ska_helpers` is not a runtime dependence, right? And `testr` is an optional runtime dependence, so I don't know where we stand on optional dependencies in this manifest.

Or do the `run` packages here relate to building in any way?

jeanconn: I thought ska_helpers was a runtime dependence if you want to guarantee you can get the version (with ska_helpers.get_version) from an environment where this is installed.

jeanconn: And I didn't consider "ability to run tests" to be optional, so I thought this was a defect in the original.

taldcroft: OK, you're right. Sounds fine.

taldcroft (approved):

javierggt:
For the record. I would not approve this PR as is, because it changes the conda package name but all dependencies are still there.

jeanconn:
Technically, they all still work because they are tested as working with the previous ska.sun (which we have not removed), but sure, we can put the dependency changes in this pr.

pkg_defs/ska_sun/meta.yaml
```diff
@@ -20,16 +20,20 @@ requirements:
...
 
 test:
   imports:
     - Ska.Sun
```
jeanconn, line 35: For this, I don't think we do the conda import tests but I left it with both Ska.Sun and ska_sun here.

pkg_defs/ska_sun/meta.yaml
```diff
@@ -20,16 +20,20 @@ requirements:
...
+    - ska_sun
 
 about:
-  home: https://github.com/sot/Ska.Sun
+  home: https://github.com/sot/ska_sun
```
taldcroft, line 39: Out of scope, but maybe https://sot.github.io/ska_sun/ would be a better `home`?  I'm pleased to see that the new flat package name just works, though of course the docs are outdated.

taldcroft: And actually maybe I'm disagreeing with myself already. Perhaps the git repo is the best home instead of the docs. I don't know.

jeanconn: I think https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html suggests doc_url for that, but hard to know how much we want to use the conda conveniences and how much is already coded in skare3_tools.

pkg_defs/ska_sun/meta.yaml
javierggt: well, if it is not used, then why leave it there?

jeanconn: Fair point.  Is it used?  It used to be that the simple "import test" was part of the conda build but I only saw it when building packages locally/manually.  For the current builds are we actually skipping it or am I just not seeing it?

There could be some reason to keep it around for the specific case of local build testing and to somewhat "announce" that this package provides both imports.

javierggt (approved):

PR #983, opened 2022-12-24
Add project template files

taldcroft:
This adds a new directory `templates` which provides baseline versions of project files. 

The immediate driver is for new files related to [applying black and isort](https://github.com/sot/skare3/wiki/Apply-black-and-isort-autoformatting-and-precommit) so that the file contents don't have be copied from a wiki. 

This might be useful in a broader context, but I'm also open to narrowing the scope to just that application and renaming the `templates` directory or maybe making it `templates/black-isort/..`. The question is whether @javierggt or @jeanconn would see utility in a general purpose place for project template files. Although we don't make new packages very often, having a template `setup.py` would be a good thing.


javierggt:
I added this PR: https://github.com/sot/skare3/pull/984 to merge on this branch (or I could make the changes in this branch, whatever you prefer).

javierggt:
Do we also want our custom workflow templates? ([docs](https://github.com/sot/Quaternion/blob/master/.github/workflows/docs.yml), [build](https://github.com/sot/Quaternion/blob/master/.github/workflows/build.yml), [release](https://github.com/sot/Quaternion/blob/master/.github/workflows/release.yml))

The only thing is they might get outdated. I keep thinking of changing the build workflow so builds of master is uploaded to the master channel, and builds of releases to the test channel.

taldcroft:
@javierggt - ready for review again

javierggt (approved):

PR #984, opened 2023-01-03
add files for scm version

javierggt:
Since we are adding files to a package template, we might also add these, which are used for SCM versions.


taldcroft:
Great idea.

PR #999, opened 2023-01-04
Add testr to chandra_time build dependencies

javierggt:
It seems we missed this one, so the build fails.


jeanconn:
Good point, I think we were going back and forth on the cmdClass bit for each of these packages, but I shouldn't have approved the other PR without this in there (or the cmdClass thing removed).

jeanconn (approved):

PR #1009, opened 2023-01-20
De-namespace the rest of the namespace recipes

jeanconn:
De-namespace the rest of the namespace recipes


jeanconn:
I'm not sure I have cmd_states right where its repo didn't match the package name to start.

javierggt (approved):
Looks good to me. I checked that there were no stray dependences other than in ska3-* packages.

There is still the question of whether Tom will change the chandra_maneuver name to agree with this PR.

jeanconn:
Thanks!  I also don't think we have any build dependencies that will get us in trouble (there's nothing circular as these are just run dependencies).  If you think this is on the right track, I think it could make sense to merge and then we can troubleshoot any build issues (like a `-` vs `_` name disagreement) as we try to build the packages.

PR #1010, opened 2023-01-21
Remove more unused and deprecated dependencies from recipes

jeanconn:
Remove more unused and deprecated dependencies from recipes

### DIFF
--- a/pkg_defs/acis_thermal_check/meta.yaml
+++ b/pkg_defs/acis_thermal_check/meta.yaml
@@ -34,17 +34,13 @@ requirements:
     - docutils
     - astropy
     - backstop_history
-    - chandra_cmd_states
-    - chandra_time
     - kadi
     - ska_dbi
     - cheta
     - ska_file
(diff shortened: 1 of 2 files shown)


javierggt (approved):
Looks fine by me, especially considering we do not care much about these dependencies.

Strictly speaking, if we list the dependencies then chandra_time would be replaced by cxotime, and the others?

jeanconn:
Right, but I checked acis_thermal_check, for example, and it still had calls to Chandra.Time, so I left it alone.

javierggt:
> I checked acis_thermal_check, for example, and it still had calls to Chandra.Time, so I left it alone

what do you mean? you removed the dependency from acis_thermal_check, and I think that's fine: I searched and I did see any use of Chandra.Time. So we might be talking of two different things :-)

jeanconn:
Sorry, yeah, it was the other way.. with chandra_time calls in the kadi code.

PR #1086, opened 2023-04-04
Make a process to install non-fsds packages

jeanconn:
This PR is at the "brainstorming" state.


jeanconn:
@javierggt @taldcroft what are your thoughts on moving these packages out of ska3-flight?  Is jobwatch fair game (it watches but does not run flight tasks)?

taldcroft:
We need to be a little careful about pulling stuff out of FSDS control. E.g. `kalman_watch` is generating email alerts for low-kalman intervals that are potentially actionable, so that counts as functionality that should be controlled. This line might be a little fuzzy so we'll need to discuss for each package and also get FSDS approval to delegate control to ACA.

jeanconn:
Right.  Though I figure all trending can have impacts that are _potentially_ actionable, so I figure we want to talk about this more and of course would need FSDS review on removing any package from explicit FSDS control.

jeanconn:
Are we thinking to leave this as ska3-non-fsds or leaning toward something like ska3-aca-pcad?  I figured keep the "ska3-" as it might make sense for logic in package building, but even that is not cast in stone.

taldcroft:
Not all of them are (or will be) relevant to PCAD. Jobwatch is an immediate example. As for the name, I still don't really like `ska3-non-fsds`, so what about:

- `ska3-tools`
- `ska3-ground` (the counter to `flight`)

jeanconn:
Who would approve an update to jobwatch?

javierggt:
The proposal you made was to have packages that are approved just by PCAD and ACA, so perhaps we can call it `ska3-pcad`?

Then the question is whether there are packages that need to be approved by other people (e.g. jobwatch)

taldcroft:
`jobwatch` would be controlled just by the Ska team (which by coincidence overlaps with the ACA team). I think we don't want to encode the control authority in the meta package name, but delegated-from-FSDS is really the defining characteristic: control of these packages is delegated to a "lower level control board" (LLCB) and can be updated and installed with approval of the lower level board. That LLCB might be ACA + PCAD, or Ska/ACA, or maybe something else in the future.

taldcroft:
A checklist would work. I'd like to keep the number of meta-packages to a minimum to reduce complexity.

taldcroft:
I spent a few minutes thinking and did not come up with anything better than ska3-non-fsds, so let's just run with that.

jeanconn:
I cut this down to just the list I think make sense for a first release.  @javierggt should this package have an explicit release baked into the meta.yaml or do we want to handle that with some github scm magic?  I'm thinking about a future where we likely want to automatically install the latest release (not master) of each of the packages in this list on HEAD.

javierggt:
Every package has the release written in meta.yml

javierggt:
@jeanconn 
> Our non-meta packages use "{{ SKA_PKG_VERSION }}"

Yes, because those are usually python packages that get their version from version control, but this is not possible with meta-packages. That is why all meta-packages have their version set in yaml. It is where the package is defined, as opposed to python packages, where the version is defined by the source code.

@taldcroft 

> Will something like this work?

That is how it would work in an ideal world, except I do not see why we would have another conda channel. The usual channel should work. What is controlled is the ska3-flight package, which has versions pinned, right? Not the channel itself.

Also, the process of creating a new version of `meta.yml` is not as easy, because it can require packages from different channels (e.g. astropy) and installed in a specific order. That is not nice, but the order can affect the final list of packages. One can start with the idea, but it might get complicated. What I did in [ska3-core-latest](https://github.com/sot/skare3/tree/master/pkg_defs/ska3-core-latest) and [ska3-flight-latest](https://github.com/sot/skare3/tree/master/pkg_defs/ska3-flight-latest) was to have a script that perhaps could be generalized, but right now it is not because we have only two cases and two different scripts, so generalizing it would complicate things. The one in ska3-core depends on a sequence of environment files.

Then, to make the final `meta.yml` one has to decide what goes in it (i.e. "external" packages should be in ska3-core while internal should be in ska3-flight). You might say that in this case that's very easy, and I agree, but then I would counter that if it is that easy, then it is not necessary to have an automated generation at all, and manually editing `meta.yml` is enough.

jeanconn:
Regarding:

> @jeanconn
> 
> Our non-meta packages use "{{ SKA_PKG_VERSION }}"
> 
> Yes, because those are usually python packages that get their version from version control, but this is not possible with meta-packages. That is why all meta-packages have their version set in yaml. It is where the package is defined, as opposed to python packages, where the version is defined by the source code.

Yes, that's why I was wondering if we wanted to try to version this non-fsds stuff differently or just not version it at all.  I was leaning toward moving toward an environment.yaml file with just these non-fsds packages listed in it without versions.  Which sounds like it is where Tom ended up too.

Regarding:

> That is how it would work in an ideal world, except I do not see why we would have another conda channel. 

Right, now -- Packages get conda packages in masters on every push to masters/main.  Packages get conda packages built in test for every github release.  Javier promotes packages to flight for flight releases.   So then what to do with these non-fsds packages?  Do we need a promotion process where they end up in the flight channel?  Or would it be fine to install them out of the test channel?  I think it would be fine to install them out of the test channel and have no "promotion" process.  Honestly, the test channel should really just be called the 'releases' channel.

But I think we're not on the same page in your last paragraphs Javier because I think Tom and I were thinking something like not replacing the ska3-flight metapackage infrastructure but adding a non-fsds.yml file.  And it would be the expectation that these non-fsds packages don't require packages not already installed from ska3-core or ska3-flight.

taldcroft (approved):
Looks good!

PR #1136, opened 2023-07-23
Add vv_trending recipe

jeanconn:
Add vv_trending recipe


jeanconn:
I confirmed in https://github.com/sot/vv_trending/pull/5 that the built package worked.

taldcroft (approved):
Looks good.

PR #1138, opened 2023-07-23
Add fid_drift_mon recipe

jeanconn:
Add fid_drift_mon recipe


jeanconn:
I've confirmed this recipe makes a conda package.  I think that is sufficient for recipe testing.

PR #1139, opened 2023-08-02
Add fss_check recipe

jeanconn:
Add fss_check recipe


jeanconn:
I tested this with a local / ska_builder build and confirm it made a package.  I did not install and test that package.

taldcroft (approved):
Looks good.

PR #1141, opened 2023-08-02
Update default ska_builder python and numpy versions

jeanconn:
Update default ska_builder python to 3.10 and numpy to 1.21

I think this is only used when building manually.

### DIFF
--- a/ska_builder.py
+++ b/ska_builder.py
@@ -47,8 +47,8 @@ def get_opt():
                         action="store_true",
                         help="Build only architecture-specific packages")
     parser.add_argument("--python",
-                        default="3.8",
-                        help="Target version of Python (default=3.8)")
+                        default="3.10",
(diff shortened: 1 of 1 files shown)


javierggt:
Good point. Now that you mention it... the numpy version should be changed to.

For reference, the current build workflow specify them explicitly: https://github.com/sot/skare3/blob/master/.github/workflows/package_build.yml

The numpy version there is 1.21

jeanconn:
Ah.  OK I'll update numpy too.  Though that makes me wonder if we should have the script and the workflow both read from something in the repo.

javierggt:
That complicates the workflow. I would leave it.

jeanconn:
OK.  I just updated the numpy version too for this.

javierggt (approved):

Issue #1146, opened 2023-08-07
Tracking new packages, recipes and non-FSDS migration


taldcroft:
## Description

A number of current ska3-flight packages are used by a well-defined subset of the CXC operations community. In order to facilitate more agile development and deployment of package updates, we propose putting these packages in a new control process which requires review and approval by only the identified stakeholders. This can be done by email.

## Package manifest

| Package | Stakeholders |
| --- | --- |
| aca_hi_bgd | ACA, PCAD |
| aca_weekly_report | ACA |
| acdc | ACA, PCAD |
| aimpoint_mon | ACA, PCAD |
| annie | ACA |
| astromon | ACA, PCAD |
| attitude_error_mon | ACA, PCAD |
| fid_drift_mon | ACA, PCAD |
| fss_check | ACA, PCAD |
| jobwatch | ACA |
| kalman_watch | ACA, PCAD |
| obc_rate_noise_plots | ACA, PCAD |
| perigee_health_plots | ACA, PCAD |
| ska_report_ranges | ACA |
| vv_trending | ACA |
| twiki-wg | ACA |

```[tasklist]
### Tasks
- [ ] https://github.com/sot/skare3/pull/1145
- [ ] https://github.com/sot/twiki-wg/pull/24
- [ ] https://github.com/sot/fid_drift_mon/pull/15
- [ ] https://github.com/sot/fid_drift_mon/pull/16
- [ ] https://github.com/sot/skare3/pull/1138
- [ ] https://github.com/sot/attitude_error_mon/pull/16
- [ ] https://github.com/sot/skare3/pull/1137
- [ ] https://github.com/sot/fss_check/pull/16
- [ ] https://github.com/sot/fss_check/pull/21
- [ ] https://github.com/sot/skare3/pull/1139
- [ ] https://github.com/sot/vv_trending/pull/5
- [ ] https://github.com/sot/skare3/pull/1136
- [ ] https://github.com/sot/skare3/pull/1153
- [ ] https://github.com/sot/attitude_error_mon/pull/18
- [ ] https://github.com/sot/twiki-wg/pull/25
- [ ] https://github.com/sot/obc_rate_noise_plots/issues/4
- [x] Promote ska3 aimpoint_mon
```

```[tasklist]
### Migrated to non-fsds process
- [x] aca_hi_bgd
- [x] aca_weekly_report
- [x] acdc
- [x] aimpoint_mon
- [x] annie
- [x] astromon
- [x] attitude_error_mon
[...]

PR #1153, opened 2023-08-16
Remove trending packages from ska3-flight/latest

jeanconn:
Remove trending packages from ska3-flight/latest


jeanconn:
@javierggt @taldcroft What are your thoughts on this list to remove?  I note that compared to https://github.com/sot/skare3/issues/1146 this *leaves* aca_view and find_attitude as part of the metapackage as I thought that might be the best plan for both ska3-flight and ska3-matlab.

One thing I note is that acdc has that email code that is called by other code -- but it is called by other code in the trending packages so that seems not so bad.  It would be better if that email code still ended up in ska_helpers.  Similarly a question about if that scaling code should end up in chandra_aca so that if we remove acdc from ska3-flight/matlab we still have the updated scaling available.

pkg_defs/ska3-flight-latest/meta.yaml
```diff
@@ -30,17 +23,13 @@ requirements:
...
     - kadi
-    - kalman_watch
     - maude
     - mica
     - parse_cm
-    - perigee_health_plots
     - proseco
     - pyyaks
     - quaternion
-    - ska-sphinx-theme
```
javierggt, line 43: This one is the sphinx theme, which is required to build the docs, so I prefer not to remove it. It has not changed in more than three years, I think.

If we remove this, the workflow has to be changed to add it, and users should add it manually before trying to build docs.

jeanconn: Sounds like a good reason to just leave it.

taldcroft (approved):

PR #1168, opened 2023-08-22
Add simple meta/recipe for chandra_limits

jeanconn:
Add simple meta/recipe for chandra_limits.  

I figured this could be tested with https://github.com/sot/chandra_limits/pull/1 , though I have not done so.  And https://github.com/sot/chandra_limits/pull/1 would still need the workflows for the actions for this to do something on merge.

### DIFF
--- a/pkg_defs/chandra_limits/meta.yaml
+++ b/pkg_defs/chandra_limits/meta.yaml
@@ -1,0 +1,39 @@
+package:
+  name: chandra_limits
+  version:  {{ SKA_PKG_VERSION }}
+
+build:
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+
+source:
+  path: {{ SKA_TOP_SRC_DIR }}/chandra_limits
+
+
+# the build and runtime requirements. Dependencies of these requirements
+# are included automatically.
+requirements:
+  # Packages required to build the package. python and numpy must be
+  # listed explicitly if they are required.
+  build:
+    - python
+    - setuptools
+    - setuptools_scm
+    - setuptools_scm_git_archive
+  # Packages required to run the package. These are the dependencies that
+  # will be installed automatically whenever the package is installed.
+  run:
+    - python
+    - numpy
+    - pytest
+    - testr
+    - ska_helpers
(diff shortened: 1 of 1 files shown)


jzuhone:
@jeanconn @javierggt let me know what I need to add re: workflows and I can do it. it's not an acisops package per se, so it probably needs whatever standard workflows sot has (if they are different at all).

jeanconn:
Thanks @jzuhone !  I think the updated workflows I stuck in https://github.com/jzuhone/chandra_limits/pull/1 might do it but @javierggt is more expert.

jeanconn:
Yes, I wasn't sure if anything used the "org" to determine this or if it was just the workflow selection.  I think in this case we do want a new package def / recipe in sot/skare3 (which is this PR) and have the workflows in sot/chandra_limits reference it.

jeanconn:
I don't see an easy way to test this against the dev branch, so I suggest just merging and forward and fix as needed.

javierggt (approved):

jzuhone:
@jeanconn @javierggt we have a first release of `chandra_limits`: https://github.com/sot/chandra_limits/releases/tag/0.5.0

jeanconn:
Thanks @jzuhone!  With the release, a conda package was successfully built and put in the test channel.  So, a few questions:

1) Does chandra_limits need to be a private repo?
2) What is the plan with management of it?  Does it belong in ska3-flight as something that we update with regular ska3-flight releases?
3) Not really a question but looks like it is still failing ruff and black in the release.

jzuhone:
Hi @jeanconn,

> 1. Does chandra_limits need to be a private repo?

No, I don't think so. I wasn't even aware that it was. @matthewdahmer can you comment? 

> 2. What is the plan with management of it? Does it belong in ska3-flight as something that we update with regular ska3-flight releases?

Yes, it needs to go in ska3-flight, as `acis_thermal_check` (among other things) are going to depend on it. 

> 3. Not really a question but looks like it is still failing ruff and black in the release.

Yes, I plan on addressing this soon.

PR #1176, opened 2023-09-18
Update pyproject.toml to configure pylance behavior in VS code

taldcroft:
## Description

This might be controversial, but I often see many code warnings from pylance that come from half-baked notebooks that I'm working on within a code repo. These warnings obscure real warnings from production code. This update to `pyproject.toml` fixes that so that pylance warnings never appear for ipynb notebooks.

Annoyingly, there is a VS code setting that *should* work, but it doesn't:

image

This might stem from the odd difference between `exclude` and `ignore` in pyright (which is called by pylance):
https://github.com/microsoft/pyright/blob/main/docs/configuration.md#main-configuration-options

## Testing

I used this in the `agasc` repo and VS code stopped showing warnings from a notebook in my editor.

### DIFF
--- a/templates/pyproject.toml
+++ b/templates/pyproject.toml
@@ -18,3 +18,11 @@ exclude = '''
 
 [tool.isort]
 profile = "black"
+
+[tool.pyright]
+ignore = [
+    "**/node_modules",
+    "**/__pycache__",
+    "**/*.ipynb",
+    ".git"
+]


javierggt (approved):
I don't know which warnings are the ones you care about, but I think it's ok to be a more lenient with notebooks.

PR #1188, opened 2023-09-20
Remove acis_taco from non-fsds package list

jeanconn:
Remove acis_taco from non-fsds package list

Our plan is to manage this non-fsds package list from master so this doesn't necessarily need to be in 2023.7 .


javierggt:
right. I will just add this to the other branch

taldcroft (approved):
Sorry for not thinking of this dependency (but go team for community review!).

PR #1222, opened 2023-12-29
Changes in local skare3 for osx-arm64

jeanconn:
Changes in local skare3 for osx-arm64


jeanconn:
This is mostly superseded by #1290 #1288 and #1289

PR #1253, opened 2024-02-08
correct ruff version in pre-commit config

javierggt:
This should have been in #1252


jeanconn (approved):
I don't think this is controversial.

PR #1220, opened 2023-12-20
2024.0

javierggt:
```
version=2024.0rc9
conda create -y -n ska3-flight-${version} --override-channels \
  -c https://ska:${CONDA_PASSWORD}@cxc.cfa.harvard.edu/mta/ASPECT/ska3-conda/test \
python==3.11.7
conda activate ska3-flight-${version}
```
```
conda install -y --override-channels \
  -c https://ska:${CONDA_PASSWORD}@cxc.cfa.harvard.edu/mta/ASPECT/ska3-conda/test \
  mamba
```
```
mamba install -y --override-channels \
  -c https://ska:${CONDA_PASSWORD}@cxc.cfa.harvard.edu/mta/ASPECT/ska3-conda/test \
  ska3-core==${version}   ska3-flight==${version}   ska3-perl==${version}
```

Actually I prefer:
```
version=2024.0rc9
conda create -y -n ska3-flight-${version}
conda activate ska3-flight-${version}
echo channels: > $CONDA_PREFIX/condarc
echo "  - https://ska:ska-cxc-20y@cxc.cfa.harvard.edu/mta/ASPECT/ska3-conda/test" >> $CONDA_PREFIX/condarc
conda install -y mamba
mamba install -y ska3-core==${version}   ska3-flight==${version}   ska3-perl==${version}
```

PR #1264, opened 2024-02-15
Python 3.11 build (speedy)

### LINKED ISSUE #1203: Consider using pip for conda build command in recipes
I ran into a problem with a development environment where I wanted to install dev versions of a few packages. I forgot that one needs to use `python setup.py install --single-version-externally-managed ..` and instead used the natural `pip install .`. This does not work for our namespace packages because pip gets confused and removes the entire namespace directory (e.g. `Chandra`) so installing `chandra_maneuver` removes `Chandra.Time` and things break.

However if you do all the installing consistently with pip then everything works as expected. This is documented here:

https://github.com/taldcroft/sample-namespace-packages/pull/1

The question is whether `pip install .` will work for all our conda recipes. You would think so given that running `python setup.py` directly is strongly deprecated now. We should consider this for ska3 2024.1.

I don't remember precisely what didn't work in the past, but now with native namespace packages hopefully it all just works. 🤞 I did confirm that it works to install all of `chandra_maneuver`, `chandra_time` and `cmd_states` packages with pip to a ska dev env. The first time when it was overriding the conda package then it removed the others, but after that everything went smoothly.

javierggt:
These are updates/improvements to the recipes and workflows for python 3.11. I used these to build python 3.11 packages from scratch. 

Normally, I do not include workflows in PRs, and do it directly instead. **This PR includes changes to recipes**.

Fixes #1203

## Updatest to recipes:
- build recipes use `pip install` instead of `python setup.py install`.
- add setuptools_scm_git_archive to build requirements wherever it was needed.
- `ska_builder.py` now passes arguments to `conda build`. This is so the build environment that is created by `conda build` can have our packages (`testr` and `ska_helpers`).
- Use `python -m setuptools_scm` instead of `python setup.py --version` in `ska_builder`. This finally gets rid of the PEP517 warning.

## Updates to workflows:

package_build workflow:
- use the latest setup-miniconda
- add optional channel to build workflow
  - use channel to create top-level environment
  - use channel in build environment (passing arguments to conda build)
- install ska_helpers and testr in top-level environment (only if they are in the conda channel)
- fix so build does not run on Mac and Windows unless noarch==False
- cache conda environment and packages
- cache conda env using hardwired directory names (a workaround)

build_package workflow:
- specify python and numpy versions
- fix the conditional to upload
- Make it so it writes a summary in the workflow summary page
- add arguments so it can be used to build packages with different python versions
- add arguments so packages can be uploaded to any conda channel (test, speedy, etc)

conda_meta.yml workflow:
- use the latest setup-miniconda
- change conda channels in environment

# Functional Tests

These workflows were used to produce the 2024.0 pre-releases. Yesterday and today I made some updates to improve documentation, and to make sure I can use the same workflow to build in different python versions, and using different conda channels.

To test these workflows, one needs the main workflow in master, so I have a copy of the `build_package.yml` workflow in another repo (test-actions), and there I ran the following:
```
from skare3_tools import github
repo = github.Repository("sot/test-actions")

repo.dispatch_event(
   "build-package",
   client_payload={
       "package": "sot/cxotime",
       "tag": '3.8.0',
       "arch": False,
       "skare3_branch": "speedy-build",
       "python_version": "3.10.8",
       "channel_in": "flight",
       "channel_out": "test",
       "upload": False,
   }
)

repo.dispatch_event(
   "build-package",
   client_payload={
       "package": "sot/cxotime",
       "tag": '3.8.0',
       "arch": False,
       "skare3_branch": "speedy-build",
       "python_version": "3.11",
       "channel_in": "speedy",
       "channel_out": "speedy",
       "upload": False,
   }
)
```

and these are the corresponding workflows:
- https://github.com/sot/test-actions/actions/runs/7935314705
- https://github.com/sot/test-actions/actions/runs/7935311003

### DIFF
--- a/.github/workflows/conda_meta.yml
+++ b/.github/workflows/conda_meta.yml
@@ -14,11 +14,14 @@ jobs:
       matrix:
         os: ["ubuntu", "macos", "windows"]
     steps:
-      - uses: sot/setup-miniconda@v2
+      - name: Setup Conda Environment
+        uses: conda-incubator/setup-miniconda@v3
         with:
-          miniconda-version: "py39_4.12.0"
-          python-version: "3.9"
-          channels: defaults,conda-forge,https://ska:${{ secrets.CONDA_PASSWORD }}@cxc.cfa.harvard.edu/mta/ASPECT/ska3-conda/flight
+          auto-update-conda: true
+          use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
+          python-version: ${{ github.event.client_payload.python_version }}
+          activate-environment: conda-build
+          channels: conda-forge,https://ska:${{ secrets.CONDA_PASSWORD }}@cxc.cfa.harvard.edu/mta/ASPECT/ska3-conda/${{ github.event.client_payload.channel }}
       - name: Checkout Skare3
         uses: actions/checkout@v2
         with:
@@ -29,7 +32,7 @@ jobs:
         shell: bash -l -e {0}
         run: |
           conda info
-          conda env update -f skare3/pkg_defs/ska3-core-latest/environment.yml
+          conda env update -f skare3/pkg_defs/ska3-core-latest/base_environment.yml
           python ./skare3/pkg_defs/ska3-core-latest/install_from_scratch.py
           conda list --json > ska3-core-${ARCH}.json
       - name: ska3-flight
--- a/.github/workflows/build_package.yml
+++ b/.github/workflows/build_package.yml
@@ -1,23 +1,132 @@
 name: Build Ska Package
 
+# This workflow builds a Ska package and optionally uploads it to a ska3-conda channel.
+
+# It takes the following arguments:
+#
+#   - package: the name of the package to build
+#   - tag: the tag to use for the package
+#   - arch: whether to build it for all architectures or as noarch (default is false)
+#   - upload: whether to upload the package to the ska3-conda channel (default is false)
+#   - python_version: the python version to use for the build (default is 3.11)
+#   - numpy_version: the numpy version to use for the build (default is 1.26.2)
+#   - channel_in: the channel to use for the build environment (default is flight).
+#     This channel gets passed to package_build workflow, which also adds conda-forge
+#   - channel_out: the channel to upload the package to (default is test)
+#   - skare3_branch: the branch of skare3 to use for the build (default is master)
+#
+# You call it like this:
+#     from skare3_tools import github
+#     repo = github.Repository("sot/skare3")
+#     repo.dispatch_event(
+#         "build-package",
+#         client_payload={
+#             "package": "sot/ska_helpers",
+#             "tag": '0.13.0',
+#             "arch": False,
+#             "upload": True
+#         }
+#     )
+
 on:
   repository_dispatch:
     types:
       - build-package
 
 jobs:
+  # this job checks the arguments, sets defaults, and writes a summary
+  arguments:
+    runs-on: ubuntu-latest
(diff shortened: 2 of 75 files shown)


.github/workflows/conda_meta.yml
```diff
@@ -29,7 +32,7 @@ jobs:
         shell: bash -l -e {0}
         run: |
           conda info
-          conda env update -f skare3/pkg_defs/ska3-core-latest/environment.yml
+          conda env update -f skare3/pkg_defs/ska3-core-latest/base_environment.yml
```
jeanconn, line 35: The file in there is a 3.10 file -  not sure if it matters or if there's an uncommited change you have hanging around which updates it.

javierggt: That file is the one used to make ska3-core from scratch. It is updated in #1220 

By the way, the changes in this PR are also in #1220. I created a branch like the 2024.0-branch, and rebased interactively to tidy it up.

jeanconn: You've lost me a bit. I thought this PR could be reviewed independently because it is against master and master has a 3.10 version of the base_environment file and there are no changes to it in this PR.  So ...

jeanconn: I thought even if this is basically a PR to capture workflow updates that it would be useful to know that all the files at speedy-build would actually work together.

jeanconn:
OK, if this PR doesn't include metapackage updates, I suppose it would work to give you a 3.11 set of packages but no meta-packages to use to install them.  If that's the goal that's fine.

javierggt:
Right. This is not touching anything related to meta-packages, this is just to build packages with newer python.

Strictly speaking `.github/workflows/conda_meta.yml` is to produce the `meta.yaml` for meta-packages, and it sneaked in this PR by virtue of being a workflow that you might not care too much about. If you think about it, this is actually a fix because the current master version of the workflow would not work, because `skare3/pkg_defs/ska3-core-latest/environment.yml` doesn't even exist.

javierggt:
I think this is still not done, because one actually should run `pip install --use-pep517 .`

I wonder why the option is needed.

javierggt:
I looked into why I was still getting the PEP517 warning, and realized it is because of the use of `python setup.py --version` in `ska_builder`. I made a quick test in my machine using cxotime as example, and just replacing that by `python -m setuptools_scm` gets rid of the warning.

Instead of the PEP517 warning:
```
/Users/javierg/SAO/miniconda3/envs/ska3-flight-2024.0rc9/lib/python3.11/site-packages/setuptools/__init__.py:80: _DeprecatedInstaller: setuptools.installer and fetch_build_eggs are deprecated.
!!

        ********************************************************************************
        Requirements should be satisfied by a PEP 517 installer.
        If you are using pip, you can try `pip install --use-pep517`.
        ********************************************************************************

!!
```

I get this:
```
Warning: could not use pyproject.toml, using default configuration.
 Reason: [Errno 2] No such file or directory: '/private/var/folders/kt/d9szn3j17wdbs3jv3vs35hcc0000gn/T/tmp1mxq_heb/cxotime/pyproject.toml'.
/Users/javierg/SAO/miniconda3/envs/ska3-flight-2024.0rc9/lib/python3.11/site-packages/setuptools_scm/_config.py:71: UserWarning: relative_to is expected to be a file, its the directory .
assuming the parent directory was passed
```

but the package built fine.

I am therefore adding that change to this PR.

jeanconn (approved):
This seems fine to me with the understanding that these updates have just been checked by using this PR locally to build the packages.

PR #1265, opened 2024-02-15
Set ska_trend to use pip install

jeanconn:
Set ska_trend to use pip install

### DIFF
--- a/pkg_defs/ska_trend/meta.yaml
+++ b/pkg_defs/ska_trend/meta.yaml
@@ -3,7 +3,7 @@ package:
   version:  {{ SKA_PKG_VERSION }}
 
 build:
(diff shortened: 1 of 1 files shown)


javierggt (approved):
simple enough. Looks good to me. We'll merge when we merge #1264

jeanconn:
Right, I suppose I could just make it a PR _against_ #1264 and merge it now?  I just didn't want new recipes to miss getting this update.

taldcroft:
This is exciting! Is it intended as a pathfinder exercise for the rest to see if there are any problems?

javierggt:
Actually, all packages with python 3.11 were built with `pip install .` already, so the speedy tests already include that.

I will add a comment to #1264 about this now.

PR #1269, opened 2024-02-25
Update ruff template redux

taldcroft:
This factors out the ruff config into a common base file `ruff-base.toml` and a project-specific file `ruff.toml`. 

The goal is to make it easier to handle ruff updates across projects by only updated `ruff-base.toml` from the `skare3/templates` area.

This is used in https://github.com/sot/sparkles/pull/205, where the `ruff-base.toml` there is exactly the same as in this PR. From a practical perspective this worked pretty well.


javierggt (approved):
I just follow your lead on this.

PR #1288, opened 2024-02-29
Update compiled recipes to use compiler directive

jeanconn:
Update compiled recipes to use compiler directive.

I confirmed that these updated recipes resulted in build packages on linux and on osx-arm64 .  I didn't not test the packages.

I did not test on windows.

These don't work to build osx-64 packages on m1-m3 macs. AFAIK the current recipes in master also don't allow for osx-64 builds from m1-m3 macs -- this update to use the compiler directive doesn't fix that lack of cross-compile support.


javierggt (approved):
looks good to me

jeanconn:
@javierggt In person I think you mentioned a testing go-back on this?  Or is it good to go?

PR #1289, opened 2024-02-29
Update ska_builder defaults and set to work from repos with master or main

jeanconn:
Update ska_builder defaults and set to work from repos with master or main.

### DIFF
--- a/ska_builder.py
+++ b/ska_builder.py
@@ -107,10 +107,12 @@ def clone_repo(name, args, src_dir, meta):
     if tag is None:
         tags = sorted(repo.tags, key=lambda t: t.commit.committed_datetime)
         repo.git.checkout(tags[-1].name)
-        if tags[-1].commit == repo.heads.master.commit:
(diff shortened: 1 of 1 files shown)


ska_builder.py
```diff
@@ -292,15 +294,14 @@ def main():
...
-        pkg_names = args.packages
+        pkg_names = sorted(args.packages)
     else:
         if args.build_list:
             with open(args.build_list) as fh:
                 pkg_names = [line.strip() for line in fh
                              if not re.match(r'\s*#', line) and line.strip()]
         else:
-            pkg_names = [str(pth.name) for pth in PKG_DEFS_PATH.glob('*') if pth.is_dir()]
-        pkg_names = sorted(pkg_names)
```
jeanconn, line 303: If you want the packages sorted for most use cases that is fine, but when building from scratch from a build_list I really don't want them sorted before building (just doesn't work), so I just put the sorting in the other blocks.

javierggt (approved):

PR #1320, opened 2024-04-02
Obc rate noise plots

javierggt:
This adds a recipe to build obc_rate_noise_plots.

Functional testing was done for sot/obc_rate_noise_plots/pull/5

### DIFF
--- a/environment-non-fsds.yml
+++ b/environment-non-fsds.yml
@@ -19,4 +19,5 @@ dependencies:
   - skare3_tools
   - vv_trending
   - twiki-wg
+  - obc_rate_noise_plots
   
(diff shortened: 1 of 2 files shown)


environment-non-fsds.yml
```diff
@@ -19,4 +19,5 @@ dependencies:
   - skare3_tools
   - vv_trending
   - twiki-wg
+  - obc_rate_noise_plots
```
jeanconn, line 22: Newline?

jeanconn (approved):

environment-non-fsds.yml
javierggt: this already had two extra lines at the end, so the file still has a newline at the end.

PR #1341, opened 2024-04-29
2024.4

javierggt:
# ska3-flight 2024.4

This PR includes:
- **acis_thermal_check**
  -  now uses chandra_limits package (acis_thermal_check [PR 71](https://github.com/acisops/acis_thermal_check/pull/71))
  - first version of 1DPAMYT model (acis_thermal_check [PR 72](https://github.com/acisops/acis_thermal_check/pull/72))
- **cheta**.
  - updated HRMA OHRTHR data to use WIDE mode.
  - Fix `Computed_Quat` MSID in cases when there are bad data in the fetch range.
- **kadi**. Remove commands v1.
- **proseco**. Implement acquisition box overlap penalty

## Interface Impacts:

- Commands v1 has been deprecated for a long time and now it is fully removed. The configuration setting commands_version is removed and the environment variable KADI_COMMANDS_VERSION no longer has any effect.
- Proseco will change which acquisition stars are selected or modify box sizes in cases where an overlapping box would have been chosen. While this code will become operation for load checking, the corresponding changes that affect load creation will not become operational until Matlab 2024_030 is promoted. As a result, there might be some discrepancies between star selection and load checking until then.
- acis_thermal_check
  - The 1DPAMYT model will have a calling sequence similar to 1DPAMZT
  - The table reporting violations on the webpages has changed slightly, but is human-read. Plots show limit lines as a function of time instead of every limit line across the whole plot; also human-read.

## Testing:

- [HEAD](https://icxc.cfa.harvard.edu/aspect/skare3/testr/releases/2024.4rc3-HEAD/).
- [OSX](https://icxc.cfa.harvard.edu/aspect/skare3/testr/releases/2024.4rc3-OSX/).
- [GRETA](https://icxc.cfa.harvard.edu/aspect/skare3/testr/releases/2024.4rc3-GRETA/).

skare3 dashboard and test result password at https://icxc.cfa.harvard.edu/aspect/skare3_dash_cred.txt

The latest release candidates will be installed in `/proj/sot/ska3/test` on HEAD, and all release candidates will be available for testing from the usual channels:
```
conda create -n ska3-flight-2024.4rc# --override-channels \
  -c https://icxc.cfa.harvard.edu/aspect/ska3-conda/flight \
  -c https://icxc.cfa.harvard.edu/aspect/ska3-conda/test \
  ska3-flight==2024.4rc#
```

## Review

[...]

PR #1342, opened 2024-04-29
Set osx libblas to be openblas not mkl

jeanconn:
On my osx_x64 installation I installed this PR in a new environment with

```
conda install ska3-core --override-channels -c file://Users/jean/git/skare3/builds/ -c https://icxc.cfa.harvard.edu/aspect/ska3-conda/flight -c conda-forge
``` 

and I end up with these packages from conda-forge

```
(new-core)skare3 jean$ conda list |grep conda-forge
libblas                   3.9.0           22_osx64_openblas    conda-forge
libcblas                  3.9.0           22_osx64_openblas    conda-forge
liblapack                 3.9.0           22_osx64_openblas    conda-forge
libopenblas               0.3.27          openmp_hfef2a42_0    conda-forge
```
And then I don't get the warning:

```
(new-core) JKRISTOFF-L1:skare3 jean$ ipython
Python 3.11.8 | packaged by conda-forge | (main, Feb 16 2024, 20:51:20) [Clang 16.0.6 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.22.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy

In [2]: 
``` 

As compared to my old(er) flight installation
```
(ska3-flight-2024.1rc4) flame:~ jean$ ipython
Python 3.11.8 | packaged by conda-forge | (main, Feb 16 2024, 20:51:20) [Clang 16.0.6 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.22.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy
Intel MKL WARNING: Support of Intel(R) Streaming SIMD Extensions 4.2 (Intel(R) SSE4.2) enabled only processors has been deprecated. Intel oneAPI Math Kernel Library 2025.0 will require Intel(R) Advanced Vector Extensions (Intel(R) AVX) instructions.
Intel MKL WARNING: Support of Intel(R) Streaming SIMD Extensions 4.2 (Intel(R) SSE4.2) enabled only processors has been deprecated. Intel oneAPI Math Kernel Library 2025.0 will require Intel(R) Advanced Vector Extensions (Intel(R) AVX) 
```

### DIFF
--- a/pkg_defs/ska3-core/meta.yaml
+++ b/pkg_defs/ska3-core/meta.yaml
@@ -1,7 +1,7 @@
 ---
 package:
   name: ska3-core
-  version: 2024.2
+  version: 2024.3
 
 requirements:
   run:
@@ -175,7 +175,8 @@ requirements:
     - lerc ==4.0.0
     - libaec ==1.1.2
     - libarchive ==3.7.2
-    - libblas ==3.9.0
+    - libblas ==3.9.0=*openblas # [osx]
+    - libblas ==3.9.0 # [linux or win]
     - libbrotlicommon ==1.1.0
     - libbrotlidec ==1.1.0
     - libbrotlienc ==1.1.0


jeanconn:
And I haven't run full testing on this potential package set.

javierggt:
So this is the only change?

If so, I can do it. It's simple enough.

The change should be in [ska3-core-latest/base_environment.yml](https://github.com/sot/skare3/blob/master/pkg_defs/ska3-core-latest/base_environment.yml#L6) and should go into 2024.3-branch

jeanconn:
Right, I wasn't as sure about how to run the multi-platform process from ska3-core-latest, so I skipped to the end with the update to ska3-core and this seems to work for me.

jeanconn:
I also wasn't sure if we wanted to go all the way to openblas on linux or if there was still some value in the mkl on that side.

javierggt:
I don't know either. I was just keeping things simple.

jeanconn:
I get you - my concern was that that changing it in ska3-core-latest without platform specifiers and doing a full core update _could_ result in bigger core package changes on all three platforms.  But I hadn't run to test.

javierggt:
and I am looking at the changes I would have to make right now, checking the diffs by hand.

jeanconn:
"and the issue is that there is no ARM build."  what does that mean?

javierggt:
> "and the issue is that there is no ARM build." what does that mean?

It means that the only reason we are doing this is so you do not get a warning. That warning only appears on arm-64, and it would go away as soon as we have an arm build.

javierggt:
I added changes for this in 2024.4. I didn’t regenerate the whole meta-package, because I did not want to wade through trivial diffs. Instead, I produced small environments with the two base environment file versions.

Windows does not change, and I believe it would be easy to only update osx.

Three packages would be removed by this comparison, but I don't know yet if other packages downstream use them:
- llvm-openmp removed on linux, not osx
- libhwloc ==2.9.3 removed on linux and osx
- tbb ==2021.11.0 removed on linux and osx

jeanconn:
Superseded by https://github.com/sot/skare3/pull/1341/commits/5bdeb586e0b6ea7e5386812e950b24fc85f759dd

PR #1365, opened 2024-06-07
Update Perl and related packages

jeanconn:


jeanconn:
@javierggt I've put some perl and core related updates for skare3 here, but I'm wondering if they can just go in a 2024.7-branch and if the packages that I've collected in /proj/sot/ska/www/ASPECT_ICXC/ska3-conda/perl could just go into the test channel.  I don't necessarily think these packages need their own skare3 PR.

javierggt:
I think we can start 2024.7-branch. That would allow me to troubleshoot the skare3 ARM build.

With these changes, do you expect ska3-perl to work on linux-64, osx-64 and osx-arm64?

jeanconn:
> With these changes, do you expect ska3-perl to work on linux-64, osx-64 and osx-arm64?

Yes, I think I've got ska3-perl straightened out, though ska3-perl-latest is certainly easier.  With regard to "work", it should install on those platforms and my initial spot checks work (starcheck looked like it ran OK). For the linux-64 packages the functional testing will need to be larger (with more testing of Replan Central, task_schedule, and splat/aca_movie).

javierggt:
I will squash these commits into one, create the branch, and test building an RC.

jeanconn:
I note the packages will likely still need to be moved to an in-use channel to work with your RC building.

javierggt:
Can you copy them to the test channel?

jeanconn:
OK.  I didn't have write permission to make a new channel at the time, but it looks like the test channel is set for kadi user so I should be able to do that.

jeanconn:
OK, copied and indexed.  I had built ska3-perl manually but I can't recall if there is a way to build that with the release process.

javierggt:
I will try to build it with the release process.

PR #1384, opened 2024-08-05
Remind about ska3-perl in PR template and Issue template

jeanconn:
Remind about ska3-perl in PR template and Issue template .

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/ska3-flight.md
+++ b/.github/PULL_REQUEST_TEMPLATE/ska3-flight.md
@@ -1,7 +1,7 @@
 # ska3-flight {version}
 
 This PR includes:
-- 
+-
 
 ## Interface Impacts:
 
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE/ska3-flight.md
```diff
@@ -16,9 +16,11 @@ The latest release candidates will be installed in `/proj/sot/ska3/test` on HEAD
 conda create -n ska3-flight-{version}rc# --override-channels \
   -c https://icxc.cfa.harvard.edu/aspect/ska3-conda/flight \
   -c https://icxc.cfa.harvard.edu/aspect/ska3-conda/test \
-  ska3-flight=={version}rc#
+  ska3-flight=={version}rc# ska3-perl=={version}rc#
```
jeanconn, line 19: For this, I suppose there's a legit process question.  We're making a new ska3-perl package for every flight release - should we just install it and save the question about "is there a perl package update?".  I'm suggesting here that we might as well.

javierggt: We rarely release ska3-perl, so most of the time this line will not work.

jeanconn: Ah.  Sorry, I thought we were just always building this - so this should be fixed.  Will do.

jeanconn:
And in my case, since I did install testing on GRETA and HEAD, starcheck did not fail - it just ran on an older-but-working version of ska3-perl.

jeanconn:
I think I've addressed your comments @javierggt .

javierggt (approved):

PR #1418, opened 2024-10-11
add aperoll

javierggt:
This PR adds a recipe for aperoll and adds it to the non-fsds environment.

I used this branch to build the latest release of aperoll, installed it from the test conda channel, and it works as expected.


taldcroft (commented):
@javierggt  - I assume you have tested installing the built package from the test channel? If so, please document that in the description. Otherwise this looks good to me.

PR #1426, opened 2024-10-29
Update ruff-base.toml to discourage using assert in code except for tests

taldcroft:
See https://docs.astral.sh/ruff/rules/assert/ and https://github.com/sot/chandra_aca/pull/174/files#r1795584783.


jeanconn (approved):
Yep.  This is what I had in mind with that comment.

PR #1436, opened 2024-11-26
release checklist step to compile all python code

javierggt:
This PR adds one step to the release checklist. After `conda update` we have to remember to run `python -m compileall`


jeanconn:
This looks like a great idea to solve the pyc issue.   I also noticed some nbc files when I was looking for not-SOT-owned files on Monday.  Is there something for numba?

javierggt:
I don't think there is. One would have to know the types of the arguments, and that is known at run time.

There might be a way to specify these types in the code, but as it is, I think the only way is to actually run the code.

jeanconn (approved):

PR #1452, opened 2024-12-18
Update template pre-commit file

jeanconn:
Update template pre-commit file

### DIFF
--- a/templates/.pre-commit-config.yaml
+++ b/templates/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.2.1
+  rev: v0.8.3
   hooks:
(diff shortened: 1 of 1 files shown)


templates/.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.2.1
+  rev: v0.8.3
```
javierggt, line 4: I just checked, and ruff 0.8.4 came out two weeks ago, so it is already outdated :)

This is bound to be outdated.

jeanconn: Yep, but being at the right major version can save some hassle if one is starting out with the templates and forgets to autoupdate.

taldcroft (approved):

PR #1471, opened 2025-01-07
set kadi requirements

javierggt:
If we want to have to requirements in the recipe, here they are.

If you prefer a subset of the requirements, we can do that too.

### DIFF
--- a/pkg_defs/kadi/meta.yaml
+++ b/pkg_defs/kadi/meta.yaml
@@ -23,36 +23,8 @@ requirements:
   # will be installed automatically whenever the package is installed.
   run:
     - python
-    - astropy
-    - django >=3.0,<4.0
-    - jinja2
-    - matplotlib
     - numpy
-    - tqdm
-    - plotly
-    - acdc
-    - agasc
-    - chandra_aca
-    - chandra_cmd_states
-    - chandra_maneuver
-    - chandra_time
-    - cheta
-    - cxotime
-    - kadi
-    - mica
-    - maude
-    - parse_cm
-    - proseco
-    - pyyaks
-    - quaternion
-    - ska_file
-    - ska_ftp
-    - ska_helpers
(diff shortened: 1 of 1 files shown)


jeanconn:
I think it isn't a _bad_ idea to update them, but since we aren't really sharing our packages and since we don't do application-specific installs, that we should probably just cut them.  

I originally found the dependencies useful to do a from-scratch build order and actually try to run tests with just the called-out dependencies--- but since we don't do that it seems like there isn't a point to trying to maintain the deps.

javierggt:
so you suggest having no dependencies at all?

jeanconn:
I was thinking 
```
  run:
    - python
```

jeanconn:
But maybe that should be:
```
  run:
    - python
    - numpy
 ```
Mostly I was thinking we should tell it that kadi needs whatever packages might be used in that conda version string solving stuff, and for the rest it just comes along with the rest of ska3 in the installed environment.

javierggt:
The python dependency is actually one of the most constraining, because it causes the package to depend exactly on the minor python version, which is why I end up having to rebuild packages for each yearly update.

I see no point in having numpy and not the others.

The only one I see some use is the django dependency, but if Tom is about to get rid of django, then that one doesn't matter either.

One argument for having the dependencies is that no packages would be missed in ska3-core when we make it again, but of course one has to maintain them.

Also, removing them from kadi will do nothing if we do not go over all packages removing them too.

jeanconn:
Oh, my mistake, I got minor version confused with bugfix version.  Minor version is right for this and I don't see how you could do anything except rebuild these arch packages for every minor version.  I'm thinking of our arch python packages as basically just tarballs that get untarred by conda install, and it is basically just putting things in "lib/python3.xx/site-packages" where 3.xx is fixed in the build.

jeanconn (approved):

PR #1479, opened 2025-01-22
Entrypoints

javierggt:
This PR sets some entrypoints explicitly. This branch was already used to build the corresponding packages, and they are already in the test conda channel. They were used to test [ska3-matlab 2025.2rc1 on windows](https://icxc.cfa.harvard.edu/aspect/skare3/testr/releases/2025.2rc1-Windows/).

To merge this, I would actually remove the commit where I set the build number.

### DIFF
--- a/pkg_defs/aca_view/meta.yaml
+++ b/pkg_defs/aca_view/meta.yaml
@@ -4,6 +4,9 @@ package:
 
 build:
   script: pip install .
+  entry_points:
+    - aca_view=aca_view.scripts.aca_viewer:main
+    - aca_view_fetch=aca_view.scripts.fetch:main
 
 source:
   path: {{ SKA_TOP_SRC_DIR }}/aca_view
--- a/pkg_defs/cheta/meta.yaml
+++ b/pkg_defs/cheta/meta.yaml
@@ -4,6 +4,14 @@ package:
 
 build:
   script: pip install .
+  entry_points:
+    - ska_fetch=cheta.get_telem:main
+    - cheta_sync=cheta.update_client_archive:main
+    - cheta_update_server_sync=cheta.update_server_sync:main
+    - cheta_update_server_archive=cheta.update_archive:main
+    - cheta_check_integrity=cheta.check_integrity:main
+    - cheta_fix_bad_values=cheta.fix_bad_values:main
+    - cheta_add_derived=cheta.add_derived:main
 
 source:
   path: {{ SKA_TOP_SRC_DIR }}/cheta
--- a/pkg_defs/cxotime/meta.yaml
+++ b/pkg_defs/cxotime/meta.yaml
@@ -4,6 +4,8 @@ package:
 
 build:
   script: pip install .
+  entry_points:
+    - cxotime=cxotime.scripts.print_time_conversions:main
 
 source:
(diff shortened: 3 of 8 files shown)


jeanconn:
I've found it is hard to know what to do with build numbers - they are handy in testing to make sure the right package is being used for testing.  But if that package with the incremented build number _is_ the one that is tested, should we just stick with the incremented build and then decrement it sometime before the next package release?  Unclear.

javierggt:
That is why I said I would remove that commit, but it helps to have it in the branch, because then one can build using the workflows. I do not know of a better way either.

jeanconn (dismissed):

taldcroft:
This looks fine, but I do find the two sources of entry points (setup.py and here) confusing (which one wins?), and of course ripe for a mismatch.

Just out of curiosity, would using `pyproject.toml` alleviate the need for entry points in the build recipe? I.e. do things work better with that?

jzuhone:
Yeah, I don't understand why the entry points need to be both here and in the package.

taldcroft:
Thanks @jeanconn. As a sanity check, I took a look at what astropy does (for pip and conda), and the answer is that they have exactly the same repetition:
https://github.com/search?q=org%3Aastropy%20astropy.io.fits.scripts.fitsdiff&type=code

So there does not appear to be a magic recipe for single-sourcing entry point scripts, so I'm satisfied that the current approach is the best we can do for now.

javierggt:
@taldcroft you asked a good question: which one wins?

I have not checked what happens in case of discrepancies. Discrepancies can be:
- an entry point missing in one of the sources.
- the same entrypoint referring to different functions in the two sources.

I think it would be good to know for sure before merging.

javierggt:
I just checked building a package with a recipe with missing entrypoints (the entrypoints being defined only in `setup.py` and not in `meta.yml`) and the resulting package did include the entrypoints, so I think it is ok to always have a minimal set of entrypoints in the recipe.

jeanconn:
Can you talk me through the last comment again?  I would think the concern would be if the entrypoint was in meta.yml and not setup.py (the inverse of the test) and did you test on Windows?

javierggt:
The question (by Tom) was

> I do find the two sources of entry points (setup.py and here) confusing (which one wins?)

My opinion is that the source code (`setup.py`) should be the ultimate authoritative source of truth.

If it is in meta.yml and not in setup.py, there is no trouble. You might get a failure which can be fixed. Even if you do not get a failure, you will end up having a non-working entrypoint that you never intended to use, because it is not in `setup.py`.

If it is in setup.py and not in meta.yml, my worry was that the entrypoint would not be created, and you would end up with a defective package (a missing entrypoint).

javierggt:
> did you test on Windows?

I tested on windows when I created this PR (that is the reason for the PR) and on windows there is no precedence issue, because it only takes the recipe.

jeanconn:
How about the case of mismatch in how the entrypoint is actually defined?  Do we get an error?  I assume the conda meta wins for the built package.

javierggt:
Yes, if the entrypoint is in both, the recipe takes precedence.

jeanconn (approved):

PR #1480, opened 2025-01-22
Update non-fsds packages environment

javierggt:
Update non-fsds packages environment:
- remove guide stat reports

### DIFF
--- a/environment-non-fsds.yml
+++ b/environment-non-fsds.yml
@@ -11,7 +11,6 @@ dependencies:
   - astromon
   - annie
   - fid_drift_mon
-  - guide_stat_reports
   - fss_check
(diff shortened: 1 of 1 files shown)


jeanconn (approved):
Yep.  Makes sense to just add it back when we're ready.

jeanconn (changes requested):
This one (find_attitude) we'd so far decided would be promoted with a non-ska3-flight fsds, but still with an fsds process.  So it would get even more confusing to put it on the non-fsds list?  Honestly we might want to just put it back in ska3-flight as the "independent fsds" idea might not work.

jeanconn (approved):

PR #1482, opened 2025-01-22
2025.3

jeanconn:


jeanconn:
I thought the outcome of the previous conversation about what to do when we had more skare3 changes for the next branch was to make the branch and start working on it, which I did.  So I would say we are already working on it. And the other two PRs I made already have base set for 2025.3-branch.  The 2023.3 PR is just a convenient place to see what is happening but maybe the one that got magically merge-closed should get deleted so it isn't confusing in the history.

jeanconn:
And the other two PRs should just get approved and merged unless you disagree with their content.

javierggt:
No. I do not disagree with the content, it is just not the process. Let me explain.

This is what the wiki page says:

> Edits to the recipes for meta-packages require a version number increase, and **are always done on a branch as part of the [release process outlined above](https://github.com/sot/skare3/wiki/Ska3-runtime-environment-for-maintainers#ska3-release-process)**: edits are made on a branch, the PR is internally reviewed, then an external authority approves. After approval, a release is made, the PR is merged and the package is installed in production environments.

and the first step in "the release process outlined above" is:

> Create a branch named {version}-branch

Those two PRs are not part of that process.

The reason for this process is to keep the meta-package dependencies and versions consistent for all meta-package `meta.yaml` files on master. In other words: if you were to look at the meta.yaml for a meta-package in the master branch, the packages listed should exactly match the ones you get if you install the meta-package with the version in the meta.yaml file.

For example: If we merge #1484, the resulting ska3-core-latest `meta.yaml` file on master will have version 2025.1, and it would have configobj listed as a dependency. However, if you install ska3-core-latest 2025.1, configobj will not be installed. That is the inconsistency the process is meant to avoid.

The two PRs can sit there until we start the release where they will be included.

jeanconn:
You've lost me because we had this conversation which is why I made them as PRs against a new 2025.3-branch and I don't understand why they can't be merged into that branch now (or when I made them).

PR #1497, opened 2025-02-11
Update ruff-base.toml to exclude "build" dir

taldcroft:
Python files hanging around in `build` should not be ruff formatted.

I tested this in `ska_trend` and confirmed that `build` was excluded as expected.


jeanconn (approved):
LGTM.

PR #1525, opened 2025-05-05
Update ska3-core libfortran for rpath issues on osx 15.4.1

jeanconn:
Update ska3-core libfortran for rpath issues on osx 15.4.1


javierggt:
That looks simple enough. Thanks for digging into this!

javierggt:
In principle, all packages for intel processors are there, but no idea if they are actually being used.

PR #1527, opened 2025-06-02
add ska3-aca

javierggt:
This PR adds a new meta-package, ska3-aca.

The ska3-aca meta-package is a superset of the ska3-flight meta-package. There are two kinds of packages in this meta-package:
- aca packages are the ones in ska3-aca and not in ska3-flight.
- flight packages are in both ska3-aca and ska3-flight, but not all patch releases of these packages will make it into a ska3-flight release. We have discussed the possibility of flight package versions differing between flight and aca, but the exact process has not been specified.

Other scripts that might need update (found doing a grep on matlab as a proxy for special cases):
- `skare3_tools/scripts/skare3_update_summary.py` and `skare3_tools/packages.py`. These scripts keep track of the versions in flight and matlab. We could add aca. If we do this, we could then change the package dashboard to show it.
- `pkg_defs/ska3-template/bin/ska_version` 
- maybe create `.github/ISSUE_TEMPLATE/release_aca.md`, the template for creating an issue for a ska3-aca-only release
- maybe create `.github/PULL_REQUEST_TEMPLATE/ska3-aca.md`, the template for creating a PR for ska3-aca-only release
- `.github/workflows/build_release.yml`. No changes required unless we want to run regression tests on ska3-aca. This build script runs regression tests after installing ska3-flight if it is being built, otherwise it will install ska3-matlab if it is built.


javierggt:
Superseded by #1536

PR #1553, opened 2025-06-20
Update ruff-base to ignore PLC0415 (imports at top)

jeanconn:
Update ruff-base to ignore PLC0415 (imports at top)


taldcroft (approved):
YES.

PR #1556, opened 2025-07-14
Update ruff-base.toml to ignore PLW1641

taldcroft:
Rule PLW1641 requires that any class that implements `__eq__` also implements `__hash__`. 

This is misguided since in practice it is common for classes to define a "liberal" view of equality where objects that compare equal do not have the same hash. As a simple example, `np.dtype("float64") == "float64"` is `True`, while the `hash()` values are not equal.

In our code base it is fine for a class object to not be hashable by default, which is what can happen by ignoring PLW1641.


jeanconn:
Makes sense to me for our use cases (and saves some noqas).

taldcroft:
I did not actually test this (that the file is well-formed). Maybe you can do it in the other PR? :smile:

PR #1557, opened 2025-07-16
Update ska3-aca

jeanconn:
Update ska3-aca with patch updates to chandra_aca and maude .


jeanconn:
@javierggt what did you decide the versions would look like for ska3-aca incremental updates?

javierggt:
I prefer continuing with the versions, unless you both prefer using the patch version. That would be consistent with what we do with ska3-flight and ska3-matlab. In this case, that would be 2025.8

javierggt:
This is in #1568

PR #1619, opened 2025-12-09
Update Sherpa recipe

javierggt:
I updated the sherpa recipe by copying the one from their repo and making some changes so it works with our script. I ran this using the package building workflow and it succeeded in building packages except on windows (as expected).

These are the differences between this PR and the sherpa recipe in the sherpa repo (not the previous version in skare3):
- use `SKA_PKG_VERSION` env variable instead of `SHERPA_VERSION` so `ska_build.py` can set it
- use `SKA_BUILD_NUMBER` instead of `SHERPA_BUILD_NUMBER`, and by default set it to `0`. We might consider using this pattern next time we need to change the build number in other packages.
- set the source using `git_url` and `git_rev` instead of a local path. This means ska_builder.py does not download the source, but conda does. We could easily change this.
- added `dev_url` to the about section
- remove explicit versions from recipe. This is better accomplished using a `conda_build_config.yaml` file, but in our case we just pass numpy and python versions in conda build.
- remove the test section. This was failing because conda build was still trying to do something, and numpy was never installed, even though I was passing the `--no-test` to conda build. Go figure.

I made some changes to `ska_builder.py`. Some are necessary to deal with this recipe:
- added a dummy macro `pin_compatible` at the top of the template. This macro just returns the package name without any version. This doesn't matter, because the dictionary that results from reading is not actually used to check the resulting pin version.
- if `dev_url` is defined in the about section, use that to fetch the sources. Up to now we have used `home` but that might not be correct (e.g: sherpa). This is used to determine the latest tag, if the `--tag` argument is not given.

The rest of the changes are improvements:
- if determining the version from setuptools_scm fails, use the version in the command-line arguments (`--tag`)
- if there is no version provided as argument, the current behavior is to check out the repo at the latest tag. After this PR, the version argument will be set to that tag, so at least it is consistent.
- if building a package fails, and there are no more packages to build after it, it doesn't ask if we "want to continue".

Note that, if the package has no tags, and the `--tag` option is not given, the script should fail (I have not checked). That is nothing new.

### DIFF
--- a/pkg_defs/sherpa/meta.yaml
+++ b/pkg_defs/sherpa/meta.yaml
@@ -8,8 +8,6 @@ source:
 
 build:
   number: {{ environ.get('SKA_BUILD_NUMBER', 0) }}
-
-build:
   skip: True  # [win]
 
 requirements:


jzuhone (approved):

javierggt:
>  we'd still need to update the defaults in ska_builder for python (and maybe numpy) etc

I just build the package as I said above, using this branch (no changes to python or numpy versions). Those changes are in 2026.0-branch.

It might be true that the exact versions are not pinned by the environment file above, so maybe that is a thing.

javierggt:
> Thanks! Have you been able to get xija_gui_fit to run on Linux using the new build?

I have not tried

javierggt:
for some reason it now fails in the workflow. Something to do with a parse error with pyproject.toml but it still works in my installation.

I say we merge this as is. We already have the packages and this is better than what is in master. We only use this once a year.

jeanconn:
Yes, I approved this with the forward and fix mindset.

PR #1663, opened 2026-03-05
Add a recipe to build replan central

jeanconn:
Add a recipe to build replan central.

This works with the build in the "package" branch in https://github.com/sot/arc/pull/96 .

Tested against ac4c84b using ska_builder.py on Linux only.

### DIFF
--- a/pkg_defs/arc/meta.yaml
+++ b/pkg_defs/arc/meta.yaml
@@ -1,5 +1,5 @@
 package:
-  name: replan_central
+  name: arc
   version:  {{ SKA_PKG_VERSION }}
 
 build:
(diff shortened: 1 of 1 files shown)


javierggt (commented):
Looks reasonable to me.

I created a package using the command documented in the skare3 wiki:
```
from skare3_tools.github import Repository

repo = Repository("sot/skare3")
repo.dispatch_event(
    event_type='build-package',
    client_payload={
        "package": "arc",
        "tag": "package",
        "arch": True,
        "upload": True,
        "skare3_branch": "replan_central",
    }
)
```

The resulting package fails to install on OSX because the package depends on packages that are only for linux (perl-extended-deps, perl-ska-convert). Was that intended?

I installed it on a test environment on HEAD and that worked. I then just checked that the entrypoints were there. No further functional check.

PR #1688, opened 2026-05-05
Add '.eggs' to extend-exclude list in ruff-base.toml

taldcroft:
Ruff is digging into an `.eggs` directory.


jeanconn (approved):
Makes sense to me.

